PR #6, opened 2016-04-10
Add cpu & memory hot add enable option

ronakbanka:
- PR add option to create vm with cpu & memory hot add enabled which can let users add cpu & memory on a running instance.

     References:
[cpu hot add enabled](https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.vm.ConfigInfo.html#cpuHotAddEnabled)
[memory hot add enabled](https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.vm.ConfigInfo.html#memoryHotAddEnabled)

- For now cpu can be increased from vsphere web client . 

Can be enabled as a resource pool property
```yaml
cloud_properties:
    cpu: 2
    ram: 4_096
    disk: 20_000
    cpu_hot_add_enabled: true
    mem_hot_add_enabled: true
```

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
@@ -1,6 +1,6 @@
 module VSphereCloud
   class VmCreator
-    def initialize(memory, disk_size_in_mb, cpu, nested_hardware_virtualization, drs_rules, client, cloud_searcher, logger, cpi, agent_env, file_provider, datacenter, cluster=nil)
+    def initialize(memory, disk_size_in_mb, cpu, cpu_hot_add_enabled, mem_hot_add_enabled, nested_hardware_virtualization, drs_rules, client, cloud_searcher, logger, cpi, agent_env, file_provider, datacenter, cluster=nil)
       @drs_rules = drs_rules
       @client = client
       @cloud_searcher = cloud_searcher
@@ -9,6 +9,8 @@ module VSphereCloud
       @memory = memory
       @disk_size_in_mb = disk_size_in_mb
       @cpu = cpu
+      @cpu_hot_add_enabled = cpu_hot_add_enabled
+      @mem_hot_add_enabled = mem_hot_add_enabled
       @nested_hardware_virtualization = nested_hardware_virtualization
       @agent_env = agent_env
       @file_provider = file_provider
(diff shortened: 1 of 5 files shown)


cppforlife:
Interesting. We'll pull it in: https://www.pivotaltracker.com/story/show/117369019

Out of curiosity when would you need this feature?

ronakbanka:
@cppforlife 
>Out of curiosity when would you need this feature?
>

sometime it is useful to increase cpu on run for debugging or incase there is high load and want to increase cpu at that moment instead of recreating the vm.

I will add another plug to this commit for `memory hot add`.

Also some future work will be needed when need to reconfigure the vm on run using bosh if these options are enabled, but that is out of scope of this PR.

ronakbanka:
@cppforlife Added the change and rebased, now this PR allows cpu & memory hot add plugs . Final for pulling in.

flawedmatrix:
Hi @ronakbanka it turns out there are some failing unit tests. See the failing check below; the details for the "concourse:unit-tests" check will bring you to a Concourse pipeline with the output from failing unit tests.

ronakbanka:
@flawedmatrix There were some missing arguments for existing tests which are added in this PR. 
Fixed.

ljfranklin:
@ronakbanka We recently finished a large refactor that made it difficult to merge this PR directly. We made the change out of band which you can see [here](https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/commit/e649e8c247f02b56e9576cbae0742d607934007b). Closing this out, thanks for the PR!

ronakbanka:
@ljfranklin I had a look at changes and want to cross check one of the changes which was there original PR and not in the master commit .

here: https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/pull/6/files#diff-d67348453c426402f004a15d36658cebR4

because if you check the original properties on vmware references they have camel case naming .

Can you confirm it the cpi changes work on your side without vmodl helper changes??

ljfranklin:
@ronakbanka That bit of code confused us at first as well, we added a [comment](https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/blob/e649e8c247f02b56e9576cbae0742d607934007b/src/vsphere_cpi/lib/ruby_vim_sdk/vmodl_helper.rb#L2) to clarify. The code does the camel casing automatically, but by default it gets confused if there are consecutive capital letters. For example, `importVApp` becomes `import_v_app` instead of the desired `import_vapp`. You only need to add an entry to that map in cases where a capital letter should not be preceeded by an underscore. The two new properties fortunately follow the normal pattern.

PR #10, opened 2016-09-28
vsphere CPI must place the "director_cid" in the agent's env, and not…

dickeyf:
… the "clean_cid".  Since encoding metadata in the CID is something specific to this CPI, the director has no clue it must remove the metedata from the CID when it talks to the agent.  The simplest way to keep the director unaware of this, would be to have the agent also use the "director_cid", so the director can talk to the agent using "director_cid" for disk operations.

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -351,6 +351,9 @@ module VSphereCloud
         end
 
         disk_spec = vm.attach_disk(disk_to_attach)
+        # Overwrite cid with the director cid
+        # Since director sends messages with "director cid" to agent, the agent needs that ID in its env, not the clean_cid
+        disk_to_attach.cid = director_disk_cid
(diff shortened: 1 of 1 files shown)


dickeyf:
A further improvement might be using <cid>.<metadata> as the disk ID everywhere...  The ID is generated by vsphere-CPI (UUID) and not by vCenter...  so would be simpler if the CPI, director, agent all used the same ID, which would happen to have metadata suffix for disk pools with datastore.

ljfranklin:
@dickeyf confirmed that this is a bug, I'll get your change merged shortly. I like the suggestion of using the "cid"."metadata" format everywhere to avoid confusion, but I remember vSphere has a character length limit of disks that wouldn't allow it.

Thanks for the heads up and PR!

ljfranklin:
@dickeyf Thanks again for the PR, once it runs through CI we'll cut a new CPI release (v29). We also added an integration test [here](https://github.com/cloudfoundry-incubator/bosh-cpi-certification/blob/56431e6ebbc77e3e6a55388a75eb2fd5f9e43c5d/vsphere/tasks/prepare-director.sh#L80) to ensure this feature functions as expected going forward.

PR #31, opened 2017-06-20
prevent javax.persistence.OptimisticLockException

sambryant4:
### What is it?

If you deploy multiple bosh directors/ deployments to the same vSphere/ NSX the auto configuring of security groups can conflict based on deployment/ job name. This causes security groups to get overwritten and can cause lock exceptions with the nsx api.

This issues is detailed under: https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/issues/28

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/nsx.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/nsx.rb
@@ -21,12 +21,12 @@ module VSphereCloud
     end
 
     def add_vm_to_security_group(security_group_name, vm_id)
+      sg_id = find_or_create_security_group(security_group_name)
       i = 0
       Bosh::Retryable.new(tries: MAX_TRIES,
                           sleep: ->(try_count, retry_exception) { 0.5 },
                           on: [Exception],
                           matching: %r{(?:javax.persistence.OptimisticLockException|<errorCode>300<\/errorCode>)}).retryer do
-        sg_id = find_or_create_security_group(security_group_name)
         if i.zero?
           @logger.debug("Adding VM '#{vm_id}' to Security Group '#{security_group_name}'...")
         else
--- a/src/vsphere_cpi/lib/cloud/vsphere/nsx.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/nsx.rb
@@ -21,32 +21,31 @@ module VSphereCloud
     end
 
     def add_vm_to_security_group(security_group_name, vm_id)
-      sg_id = find_or_create_security_group(security_group_name)
-
-      @retryer.try(MAX_TRIES) do |i|
-        if i == 0
+      i = 0
+      Bosh::Retryable.new(tries: MAX_TRIES,
+                          sleep: ->(try_count, retry_exception) { 0.5 },
+                          on: [Exception],
+                          matching: %r{(?:javax.persistence.OptimisticLockException|<errorCode>300<\/errorCode>)}).retryer do
+        sg_id = find_or_create_security_group(security_group_name)
+        if i.zero?
           @logger.debug("Adding VM '#{vm_id}' to Security Group '#{security_group_name}'...")
         else
           @logger.warn("Retrying adding VM '#{vm_id}' to Security Group '#{security_group_name}', #{i} attempts so far...")
         end
 
+        i += 1
+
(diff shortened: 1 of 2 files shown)


cppforlife:
@sambryant4 there should be a combination of sec groups that are already prefixed with a director name. any reason why not change the name so that becomes more useful in differentiating sec group use?

sambryant4:
Hi @cppforlife, thanks for the feedback, I have no issue with using the director-name prefixed groups, but at present the CPI deploys groups both prefixed and un-prefixed. Maybe a better approach would be to only deploy security groups that are prefixed with the director name?

cppforlife:
@sambryant4 our thinking was that users would just not use unprefixed security groups. any reason you want to remove them?

sambryant4:
The problem we have seen is that if you happen to be running two bosh deploys against the same NSX at the same time then NSX sometimes errors with "optimistic lock exceptions", it looks like that happens when two bosh directors are trying to update the same security group at the same time. Its also just confusing from an operations perspective where these security groups keep swapping what deployment they relate too, but I do agree that if NSX does not error we can choose to ignore them.

cppforlife:
@sambryant4 

> NSX sometimes errors with "optimistic lock exceptions"

do you happen to have error message for this. i'd like to take a look at it further and possibly fix it.

sambryant4:
@cppforlife here is an example:
```
Error: Unknown CPI error 'Unknown' with message 'Failed to add VM to Security Group with unknown NSX error: '<?xml version="1.0" encoding="UTF-8"?>
<error><details>nested exception is javax.persistence.OptimisticLockException</details><errorCode>258</errorCode></error>'' in 'create_vm' CPI method
```

sambryant4:
@cppforlife Have you had any thoughts around the above error? Happy to try different approaches in this PR if we can think of something else that might work.

cppforlife:
@sambryant4 Seems very similar to other NSX error we just fixed in v42: https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/commit/92f9bb9431e43458bbf2e96831d39a96c5c6bdd3. We can add retry logic for this error as well: https://www.pivotaltracker.com/story/show/148528415

sambryant4:
@cppforlife thanks, I'll give the retry approach a go

sambryant4:
@cppforlife Had another crack at it, does this look more like it?

cdutra:
@sambryant4 Thanks for the PR!

- We recommend moving this [line](https://github.com/FidelityInternational/bosh-vsphere-cpi-release/blob/6aa666bdfab49fa84feb3cdbd1d11c94744b3108/src/vsphere_cpi/lib/cloud/vsphere/nsx.rb#L29) outside the retrier block. We don't want to make multiple API calls to create or retrieve a security group in case of retriable error when adding a vm to the security group.
- [This](https://github.com/FidelityInternational/bosh-vsphere-cpi-release/blob/6aa666bdfab49fa84feb3cdbd1d11c94744b3108/src/vsphere_cpi/lib/cloud/vsphere/nsx.rb#L40-L47) is a bit confusing and there's room from improvement. We could log different messages for the different success cases, i.e. the case where the vm already belongs to the security group and the case where the vm is added to the security group.

cc @sshariff01

sambryant4:
@cdutra completely agree on both points, I clearly didn't eat enough cereal yesterday, updates incoming.

sambryant4:
@cdutra How does it look this time?

cdutra:
@sambryant4 Thanks! It looks great. Just one more thing: this [line](https://github.com/FidelityInternational/bosh-vsphere-cpi-release/blob/69c85a83045a2d0ff0ae68f9f4dfefa9c7d35991/src/vsphere_cpi/lib/cloud/vsphere/nsx.rb#L50) will always evaluate to true - so why not return true?

cc @sshariff01

sambryant4:
@cdutra that was an attempt to be consistent with this [line](https://github.com/FidelityInternational/bosh-vsphere-cpi-release/blob/69c85a83045a2d0ff0ae68f9f4dfefa9c7d35991/src/vsphere_cpi/lib/cloud/vsphere/nsx.rb#L50). I do agree that it seems a little excessive so will swap it out if you would rather

cdutra:
@sambryant4 Make sure to sign the CLA. Thanks!

cc @sshariff01

sambryant4:
@cdutra the CLA should be signed now! Thanks

PR #32, opened 2017-06-22
trying to present unitnum with a negative value so let vsphere to ass…

datianshi:
I tried to use the latest stemcell with latest CPI on a vSphere 6.0 environment
My use case was trying to add multiple network adaptors to a VM through bosh deployment. 
It worked fine up to 5 nics per VM, but I can not add more than 5, while I can manually add more than 5 through vSphere Client.

After extensive debugging, I suspected that the unit number was causing that the 6th NIC device unit number may be duplicate with other devices. I tried to set a temporary negative value (as the key value) and expect the server side to assign unit number. And it turned out that fixes the problem.....

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/nic.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/nic.rb
@@ -36,6 +36,7 @@ module VSphereCloud
         nic.key = -1
         nic.controller_key = controller_key
         nic.backing = backing_info
+        nic.unit_number = -1
 
         nic
       end


cppforlife:
@datianshi will take a look. we were of the opinion that we were using deprecated apis in vsphere so that's why it was limiting us.

cdutra:
@cppforlife The aforementioned commit fixes the issue while still using the 5.5 API. We tried for 10 nics and it worked. The 11th will exceed the hard limit for vSphere according to this [document](https://www.vmware.com/pdf/vsphere6/r60/vsphere-60-configuration-maximums.pdf)

cdutra:
@datianshi The actual problem is because we're assigning unit_number to VirtualEthernetCard devices when we're suppose to do this only for SCSI devices. This [commit](https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/commit/fbc274d2b6a6e550b62d35b9e928a7c9973fbb12) should fix the issue.
vSphere has a limit of 10 nics per VM.

datianshi:
@cdutra @cppforlife while this is another fix.... I still think mine is better... why add an if condition on client code to check if it is virtual network card, while the client side can be agnostic by assigning a negative value....

cdutra:
Setting `unit_number` to -1 fixes the issue because we do a check for nil on the `unit_number` when assigning those numbers.
The if condition checking for the type of the device is clearer because it makes the code explicit that this function should not be used for VirtualEthernetCard.

PR #40, opened 2017-11-09
Calculates cluster free space based on balanced score

EleanorRigby:
- Adds a method to calculate balaced score based free space to lifecycle
helpers.
- Uses new balaced score based approach to verify cluster free spaces.

### DIFF
--- a/src/vsphere_cpi/spec/integration/cluster_cloud_properties_spec.rb
+++ b/src/vsphere_cpi/spec/integration/cluster_cloud_properties_spec.rb
@@ -101,7 +101,7 @@ describe 'cloud_properties related to clusters' do
       cpi = VSphereCloud::Cloud.new(options)
       # @cluster_more_memory should have more memory than @cluster_less_memory and
       # both clusters need to have the specified @shared_datastore
-      verify_cluster_free_space(cpi, @cluster_more_datastore_free_space, @cluster_less_datastore_free_space)
+      verify_cluster_free_space_balanced_score(cpi, @cluster_more_datastore_free_space, @cluster_less_datastore_free_space)
     end
     
     let(:vm_type) do
--- a/src/vsphere_cpi/spec/support/lifecycle_helpers.rb
+++ b/src/vsphere_cpi/spec/support/lifecycle_helpers.rb
@@ -88,6 +88,24 @@ module LifecycleHelpers
 
     fail("Primary cluster, #{cluster1_name}, must have more free space (#{cluster1_free_datastore_space}) than secondary cluster, #{cluster2_name} (#{cluster2_free_datastore_space})") unless cluster1_free_datastore_space > cluster2_free_datastore_space
   end
+
+  def calculate_balance_score(cluster)
+    free_spaces = cluster.accessible_datastores.map { |_, props| props.free_space }.sort
+    min = free_spaces.first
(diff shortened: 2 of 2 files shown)


EleanorRigby:
The concern I found was in the (cluster_cloud_properties_spec.rb) tests that when we verify cluster spaces before creating a vm. 

verify_cluster_free_space merely adds all the free space. While , in process of creating a vm we take into account balanced scores for cluster selection. Both the behaviors should align, that is, verification should also be done on basis of balanced scores.

See the code snippets below 

``` 
verify_cluster_free_space(cpi, @cluster_more_datastore_free_space, @cluster_less_datastore_free_space)
 
def verify_cluster_free_space(cpi, cluster1_name, cluster2_name)
  cluster1 = cpi.datacenter.find_cluster(cluster1_name)
  cluster2 = cpi.datacenter.find_cluster(cluster2_name)

  cluster1_free_datastore_space = cluster1.accessible_datastores.map { |_, props| props.free_space }.inject(0, :+)
  cluster2_free_datastore_space = cluster2.accessible_datastores.map { |_, props| props.free_space }.inject(0, :+)

  fail("Primary cluster, #{cluster1_name}, must have more free space (#{cluster1_free_datastore_space}) than secondary cluster, #{cluster2_name} (#{cluster2_free_datastore_space})") unless cluster1_free_datastore_space > cluster2_free_datastore_space
end
```

```
 def placements_with_max_free_space(placements)
      sorted = placements.sort_by { |_, placement| placement[:balance_score] }
      maximum = sorted.last[1][:balance_score]

      sorted.select { |_, placement| placement[:balance_score] == maximum }.to_h
 end
```
 
 
 
This simply adds free space while cluster picker calculates based on balanced score, which is wrong. We should be doing same thing for verification also.

Hence, this change.

EleanorRigby:
Post this change, we can also drop below env vars.

export BOSH_VSPHERE_CPI_CLUSTER_MORE_DATASTORE_FREE_SPACE=XXXX
export BOSH_VSPHERE_CPI_CLUSTER_LESS_DATASTORE_FREE_SPACE=XXXX
The tests can reuse

export BOSH_VSPHERE_CPI_CLUSTER=XXX
export BOSH_VSPHERE_CPI_SECOND_CLUSTER=XXXX
This definitely is highly dependent on how we setup our testbeds. But lesser then env vars, simpler it is to setup infra.

cdutra:
@EleanorRigby We decided to have the environment variables to be specific because these tests run in different environments and each environment could be different in the one that has more or less free space.

EleanorRigby:
@cdutra 
Agree. The production code works by calculating balanced score and adding them to placements map in one single function. If I want to re use it , I will have to break it into two small functions. 
Does this approach looks good for me to go ahead and make changes?

cppforlife:
well if we are using prod code in the test then is it really testing anything?

EleanorRigby:
@cppforlife : I see your point. But the function for calculating balanced score is essentially the same as written in prod. The ramification of using it finally in a  series of calls that start from vm_create versus using it standalone to verify which cluster has more balanced score will be different. So, we can use same also.

I am ready to change as you folks suggest. I am neutral to both ways.

EleanorRigby:
@cppforlife : any update on this??

cppforlife:
@EleanorRigby ill defer to you all. just wanted to mention it.

PR #41, opened 2017-11-16
Network_spec is not memoized anymore through let call

EleanorRigby:
- spec/integration/drs_rules_spec failed due to IP conflict that emerged
due to memoized network_spec. This caused duplicate IP in integration
tests vms.
- New function call randomizes IP every time.
- Still there are slim chances (1/(255*255)) of IP collision. To make
it truly collision free , we can maintain an IP pool per "it" block and use
this pool to get new IP from get_network_spec

### DIFF
--- a/src/vsphere_cpi/spec/integration/drs_rules_spec.rb
+++ b/src/vsphere_cpi/spec/integration/drs_rules_spec.rb
@@ -16,18 +16,6 @@ describe 'DRS rules', drs: true do
       )
     end
 
-    let(:network_spec) do
-      {
-        'static' => {
-          'ip' => "169.254.#{rand(1..254)}.#{rand(4..254)}",
-          'netmask' => '255.255.254.0',
-          'cloud_properties' => {'name' => @vlan},
-          'default' => ['dns', 'gateway'],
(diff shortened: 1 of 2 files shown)


EleanorRigby:
Forward : Slack Discussion


I assume there must be an async call at some point in CPI code to vcenter to create a VM. So when vms are being created/powered on , they don’t have a IP yet. If I run query `vm = @client.find_vm_by_ip(ip)` in `ip_conflict_detector.rb` , it will return no vm as they are all in process of being powered on and that IP might not exist yet. 

===> If all of this happens slowly IP conflicts are not detected, 

===> But, if a testbed is fast and vms spawn quickly , this call query `vm = @client.find_vm_by_ip(ip)` in `ip_conflict_detector.rb` will start returning conflicting ip address vms

In nutshell , if  a testbed is slow we can quickly create vms with same IP address without ipconflict finding function ever getting to notice it.

ktchen14:
Seems fine to me for now. Eventually I’d prefer this just be called `network_spec` with all of the unnecessary `let(:network_spec)`s removed. This is a good step in that direction.

PR #42, opened 2017-11-16
Take custom cluster config into consideration when attaching disk

nehagjain15:


src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
```diff
@@ -27,20 +27,20 @@ def resource_pool
...
+            host_properties['datastore'],
             ensure_all: true
-          )
-          if ds['summary.accessible']
-            ds['name']
-          else
-            nil
-          end
-        end.compact
+        ).select { |datastore| datastore.accessible }
```
cdutra: .select(&:accessible)

src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
```diff
@@ -27,20 +27,20 @@ def resource_pool
...
-      def accessible_datastore_names
-        host_properties['datastore'].map do |store|
-          ds = cloud_searcher.get_properties(
-            store,
-            Vim::Datastore,
-            ['name', 'summary.accessible'],
+      def accessible_datastores
+        Datastore.build_from_client(
+            @client,
+            host_properties['datastore'],
```
cdutra: is there behavior the same as query multiple times? Why were we querying multiple types?

nehagjain15: why were we querying multiple times, not sure. But behavior is same as both call cloud_searcher.get_properties. get_properties takes care if obj is an array or a single ManagedObject. For this change we need Datastore object  with all properties instead of just 2 properties which were being queried earlier

nehagjain15:
@cdutra can you please review it again

PR #65, opened 2018-02-15
Nj 154268851 storage pods

nehagjain15:
Allow DatastoreClusters to be specified in vm_type and use that to create vm's and ephemeral disks. 
DO NOT MERGE. WIP


src/vsphere_cpi/spec/integration/cluster_cloud_properties_spec.rb
```diff
@@ -165,4 +165,93 @@
...
+            ]
+          }
+        ]
+      )
+      VSphereCloud::Cloud.new(options)
+    end
+
+    let(:disk_pool) { { 'datastores' => [@disjoint_datastore] } }
+
+    it 'should place disk into datastore that belongs to the cluster defined in cloud config' do
```
cppforlife: we should probably asset that stemcell is also copied into particular datastore as well

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -506,15 +506,38 @@ def delete_disk(raw_director_disk_cid)
...
+          original_stemcell_vm,
+          name_of_replicated_stemcell,
+          @datacenter.template_folder.mob,
+          cluster.resource_pool.mob,
+          datastore_cluster: datastore_cluster
+        )
+        @logger.info("Recommendation from Storage DRS for: #{name_of_replicated_stemcell}, Destination: #{recommendation.action.first.destination.name}")
+        #loop over all actions to pick one with reason as storagePlacement?
+        if recommendation.reason == 'storagePlacement' && recommendation.action.first.destination.class == VimSdk::Vim::Datastore
+          to_datastore = recommendation.action.first.destination
```
cppforlife: could there be other recommendations (not first)?

nehagjain15: RecommendDatastores can return multiple recommendations, the best one will be first entry in the array.

.gitignore
```diff
@@ -11,6 +11,7 @@ blobs
 .DS_Store
 .final_builds/jobs/**/*.tgz
 .final_builds/packages/**/*.tgz
+.byebug_history
```
ktchen14: This should go in your global gitignore rather than in this project specific one.

PR #67, opened 2018-02-27
Remove perfCounter. 

EleanorRigby:
Add cluster resource usage summary.

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
@@ -44,7 +44,7 @@ module VSphereCloud
         # Have to use separate mechanisms for fetching utilization depending on
         # whether we're using resource pools or raw clusters.
         if @config.resource_pool.nil?
-          @synced_free_memory = fetch_cluster_utilization(properties['host'])
+          @synced_free_memory = fetch_cluster_utilization
         else
           @synced_free_memory = fetch_resource_pool_utilization
         end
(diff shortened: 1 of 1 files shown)


src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
```diff
@@ -81,40 +81,22 @@ def select_datastores(pattern)
...
-            percent_used = Util.average_csv(counter['mem.usage.average']) / 10000
-            free_memory = ((1.0 - percent_used) * total_memory).to_i
-          else
-            logger.warn("host '#{host_properties['name']}' is missing 'mem.usage.average' (possibly recently booted), ignoring")
-          end
-
-          cluster_free_memory += free_memory
+      # @return
+      def fetch_cluster_utilization()
+        require 'pry-byebug'
```
EleanorRigby: Have to remove this. Sorry

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
```diff
@@ -81,40 +81,22 @@ def select_datastores(pattern)
...
-          if !counter['mem.usage.average'].nil?
-            percent_used = Util.average_csv(counter['mem.usage.average']) / 10000
-            free_memory = ((1.0 - percent_used) * total_memory).to_i
-          else
-            logger.warn("host '#{host_properties['name']}' is missing 'mem.usage.average' (possibly recently booted), ignoring")
-          end
-
-          cluster_free_memory += free_memory
+      # @return
+      def fetch_cluster_utilization()
```
nehagjain15: Method name should be effective_memory as thats what this is returning

cppforlife:
when does `overall_status` change? when is it green, red, etc?

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
EleanorRigby: This was earlier also named as cluster utilization. I would prefer to keep changes as minimal as possible. If we ever introduce cluster CPU usage as one of the requirements , then I can rename things differently.
What do you think?

EleanorRigby:
@cppforlife : Thanks for having a look at my PR
Overall Status represents a ManagedEntityStatus . 

gray | The status is unknown- Initially the alarms start with the Gray status.
green | The entity is OK and the cluster is healthy.
yellow | This is just a warning but potentially could lead to an alert (RED), but the cluster is still functional. 
red | The cluster has a serious problem and this sets of an alert - VSAN issues.

PR #69, opened 2018-03-01
Telemetry change

EleanorRigby:
- Removes PerfCounter from Cluster Usage calculation
- Adds a post start script to update 'config.SDDC.CPI option for VAC

### DIFF
--- a/jobs/vsphere_cpi/spec
+++ b/jobs/vsphere_cpi/spec
@@ -6,6 +6,7 @@ templates:
   cpi.json.erb: config/cpi.json
   nsx_cacert.pem.erb: config/nsx_cacert.pem
   nsxt_cacert.pem.erb: config/nsxt_cacert.pem
+  post-start.erb: bin/post-start
 
 packages:
 - iso9660wrap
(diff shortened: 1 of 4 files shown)


cppforlife:
is there a reason why we cannot set this property as part of whatever CPI call that needs to happen? current approach does not work for multi CPI since creds are only passed in via the CPI method call.

ktchen14:
@cppforlife This is to enable telemetry on vSphere which needs to only happen once globally. We could put it into something like `create_stemcell` but it seems more natural to have it be part of the CPI's `post-start`. Does the CPI not have access to the global configuration in a multi CPI scenario?

EleanorRigby:
@yeshwantbabar

cppforlife:
> Does the CPI not have access to the global configuration in a multi CPI
scenario?

it does not, because cpi may be called with different vcenters etc at
runtime.

EleanorRigby:
@cppforlife : So, as per the above comment it will only work on the vcenter on which director is deployed due to post-start . Other vcs in multi-cpi scenario will not be updated with the advanced property. 
But if we call this addition of property with any other CPI call (let's say stemcell creation) , it will happen over and over agin with each stemcell creation. Although this is idempotent but might lead to events piling up on vc side.
@yeshwantbabar : What do you say? Should we investigate more before pushing it? 

This will at least tell us majority of number of VC using CPI assuming multi-cpi is not widely used yet.

PR #72, opened 2018-03-13
Refactor Ephemeral disk selection into VmType & StoragePicker

nehagjain15:
- Remove DiskConfigFactory
- Introduce StorageList
- Refactor Tests

[#154268851](https://www.pivotaltracker.com/story/show/154268851)


src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -313,10 +304,10 @@ def create_vm(agent_id, stemcell_cid, vm_type, networks_spec, existing_disk_cids
             end
           end
 
-          if vm_type.key?('nsx') && !vm_type['nsx']['lbs'].nil?
-            security_groups = vm_type['nsx']['lbs'].map { |m| m['security_group'] }.uniq
+          unless vm_type.nsx_lbs.nil?
+            security_groups = vm_type.nsx_lbs.map { |m| m['security_group'] }.uniq
             security_groups.each { |sg| nsx.add_vm_to_security_group(sg, created_vm.mob_id) }
-            nsx.add_members_to_lbs(vm_type['nsx']['lbs'])
+            nsx.add_members_to_lbs(vm_type.nsx['lbs'])
```
ktchen14: Should we also do `vm_type.nsx_lbs` here?

src/vsphere_cpi/lib/cloud/vsphere/disk_pool.rb
```diff
@@ -1,33 +1,15 @@
...
-        hash = Hash.try_convert(entry)
-        next if hash.nil?
-        hash.key?('clusters')
-      end
-      return [] if clusters_hash.nil?
-      clusters = clusters_hash['clusters']
-      clusters.map do |cluster|
-        Resources::StoragePod.find_storage_pod(cluster.keys.first, @datacenter.mob)
-      end
+      @storage_list = storage_list || []
```
ktchen14: It's better to do `storage_list = []` in the function prototype than `storage_list || []` here.

src/vsphere_cpi/lib/cloud/vsphere/storage_list.rb
```diff
@@ -0,0 +1,28 @@
+module VSphereCloud
+  module StorageList
+
+    # storageList is a list of datastores and datastore_clusters defined in cloud_properties as 'datastores'
```
ktchen14: Let's put this comment above immediately above the `module` since that's the style used by YARD and `rdoc`.

src/vsphere_cpi/lib/cloud/vsphere/storage_list.rb
```diff
@@ -0,0 +1,28 @@
...
+    # Eg. [datastore1, clusters: [{datastore_cluster1: {}, datatore_cluster2: {}}]]
+
+    #Returns list of datastore names
+    def datastore_names
+      storage_list.select do |entry|
+        entry.is_a?(String)
```
ktchen14: Let's be friendly to Ruby's duck-typiness. Should be:

```ruby
storage_list.map { |entry| String.try_convert(entry) }.compact
```

See https://idiosyncratic-ruby.com/54-try-converting.html

src/vsphere_cpi/lib/cloud/vsphere/storage_list.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+    end
+
+    #Returns list of StoragePod objects for datastore_clusters
+    def datastore_clusters
+      clusters_hash = storage_list.find do |entry|
+        hash = Hash.try_convert(entry)
+        next if hash.nil?
+        hash.key?('clusters')
```
ktchen14: This block should be:

```ruby
Hash.try_convert(entry)&.key?('clusters')
```

src/vsphere_cpi/lib/cloud/vsphere/storage_list.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+      return [] if clusters_hash.nil?
```
ktchen14: This is one of those rare cases where an `or return []` may be justified at the end of the previous line. How do we feel about `and`/`or` in Ruby?

src/vsphere_cpi/lib/cloud/vsphere/storage_picker.rb
```diff
@@ -38,5 +38,49 @@ def choose_persistent_storage(size_in_mb, target_datastore_pattern, available_da
...
+    # choose storage from set of datastore and sdrs enabled datastore clusters
+    # using weight based algorithm on free space to choose the storage
+    # if datastore clusters are provided and none of them have sdrs enabled log an error
+
+    # @param [String] target_datastore_name
+    # @param [Hash] accessible_datastores
+    # @param [VmType] vm_type
+    def choose_ephemeral_storage(target_datastore_name, accessible_datastores, vm_type, logger)
+      #pick datastore set by cluster_picker based on ephemeral_pattern
+      storage_options = [accessible_datastores[target_datastore_name]].compact
```
ktchen14: This looks off. Should it be `Array(accessible_datastores[target_datastore_name])` instead?

src/vsphere_cpi/lib/cloud/vsphere/storage_picker.rb
```diff
@@ -38,5 +38,49 @@ def choose_persistent_storage(size_in_mb, target_datastore_pattern, available_da
...
+      end
+      datastore_names = datastore_names.compact
+      if datastore_names.empty? && vm_type.datastore_clusters.empty?
+        logger.info("Using global ephemeral disk datastore pattern: #{vm_type.datacenter.ephemeral_pattern}")
+        vm_type.datacenter.ephemeral_pattern
+      else
+        logger.info("Using datastore list: #{datastore_names.join(', ')}")
+        "^(#{datastore_names.map do |name|
+          Regexp.escape(name)
+        end.join('|')})$"
```
ktchen14: In general we should avoid doing this moving forward. This should be:

```ruby
Regexp.union(datastore_names.map { |name| /^#{Regexp.escape(name)}$/ })
```

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -15,7 +15,7 @@ def initialize(client:, cloud_searcher:, logger:, cpi:, datacenter:, agent_env:,
 
     def create(vm_config)
       cluster = vm_config.cluster
-      storage = choose_storage(vm_config)
+      storage = StoragePicker.choose_ephemeral_storage(vm_config.ephemeral_datastore_name, vm_config.cluster.accessible_datastores, vm_config.vm_type, )
```
ktchen14: There's a trailing comma in the argument list here.

src/vsphere_cpi/lib/cloud/vsphere/vm_type.rb
```diff
@@ -0,0 +1,28 @@
...
+  class VmType
+    include VSphereCloud::StorageList
+    attr_reader :datacenter, :storage_list
+
+    # @param [Resources::Datacenter] datacenter
+    # @param [Hash] cloud_properties
+    def initialize(datacenter, cloud_properties)
+      @datacenter = datacenter
+      @cloud_properties = cloud_properties
+      @storage_list = cloud_properties['datastores'] || []
```
ktchen14: `@storage_list` should be a method because it'll go out of sync with `@cloud_properties` if `@cloud_properties` is mutable

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
nehagjain15: Yes, I missed it

src/vsphere_cpi/lib/cloud/vsphere/storage_picker.rb
nehagjain15: This one lets chat tomorrow, as I was also undecided if I should directly pass the actual datastore here instead of first 2 parameters

src/vsphere_cpi/lib/cloud/vsphere/disk_pool.rb
nehagjain15: storage_list cab be nil if datastores property is not set in cloud config and in that case need to set it to empty array, so cannot do it in function definition

PR #82, opened 2018-04-06
Add vSphere 6.7 job and fault based error expectations

EleanorRigby:
Add job to run integration test on vsphere 6.7  …
Handle fault based error expectations in drs integration test .


src/vsphere_cpi/lib/ruby_vim_sdk/vm_power_on_exception.rb
```diff
@@ -0,0 +1,12 @@
+module VimSdk
+  class VMPowerOnError < StandardError
+
+    attr_reader :fault
+
+    def initialize(message, fault)
```
ktchen14: The `raise` method in Ruby is really strange in that it takes between 0 and 3 arguments and doesn't directly create an exception:

- 0 arguments: raises whatever is in `$!` (almost always the current exception being handled).
- 1 argument: if the argument responds to the `exception` method then it calls `exception` on the argument and raises that. Both subclasses (e.g. `TypeError`) and instances (e.g. `TypeError.new`) of `Exception` respond to the `exception` method. If the argument is a `String` then this raises a `RuntimeError` with the argument as the *message*.
- 2 arguments: raises the result of calling `exception` on the first argument with the second argument as the message.
- 3 arguments: raises the result of calling `exception` on the first argument with the second argument as the message and the third argument as the backtrace.

It's stylistically preferable in Ruby to use the 2 argument form of `raise` to supply a message rather than the 1 argument form (see https://github.com/bbatsov/ruby-style-guide#exception-class-messages). Therefore we can define the `initialize` method instead as:

```ruby
def initialize(fault)
  @fault = fault
end
```

So that later on we can raise this exception as:

```ruby
raise VimSdk::VMPowerOnError.new(fault), "Could not power on VM '#{vm}': #{fault.msg}"
```

You can probably go further here and define `to_s` as:

```ruby
def to_s
  "#{super}: #{fault.msg}"
end
```

So that the fault message is automatically appended and your `raise`s become:

```ruby
raise VimSdk::VMPowerOnError.new(fault), "Could not power on VM '#{vm}'"
```

src/vsphere_cpi/spec/integration/drs_rules_spec.rb
```diff
@@ -293,7 +293,9 @@
                 {'key' => 'value'}
               )
               vms.push(failed_vm)
-            end.to raise_error(/.*DRS cannot find a host to power on or migrate the virtual machine.*/)
+            end.to raise_error { |error|
+              expect(error.fault).to be_kind_of(VimSdk::Vim::Fault::RuleViolation).or be_kind_of(VimSdk::Vim::Fault::GenericDrsFault)
```
ktchen14: Since you're going to the trouble of defining your own `Exception` class then this logic probably belongs in it. Assuming your `VMPowerOnError` has a method called `unplaceable?` then you can match this with:

```ruby
expect do
...
end.to raise_error(VMPowerOnError, be_unplaceable)
```

EleanorRigby: raise_error expect a string as second argument. Are you sure it is possible to write a syntax like this. I tried and it does not work.

src/vsphere_cpi/lib/ruby_vim_sdk/vm_power_on_exception.rb
EleanorRigby: Fixed

src/vsphere_cpi/lib/ruby_vim_sdk/vm_power_on_exception.rb
```diff
@@ -0,0 +1,12 @@
+module VimSdk
```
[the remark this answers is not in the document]
EleanorRigby: Fixed

src/vsphere_cpi/lib/ruby_vim_sdk/vm_power_on_exception.rb
```diff
@@ -0,0 +1,12 @@
...
+  class VMPowerOnError < StandardError
+
```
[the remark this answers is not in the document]
EleanorRigby: Fixed

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
```diff
@@ -85,7 +85,11 @@ def power_on_vm(datacenter, vm)
 
         vms_powering_on = result.attempted
         if vms_powering_on.empty?
-          raise "Could not power on VM '#{vm}': #{result.not_attempted.map(&:fault).map(&:msg).join(', ')}"
+          # Since we are only powering on single vm, the not_attempted member
+          # of result will have length equal to one. Thus getting first one
+          # from not_attempted list is the right thing to do.
+          power_on_exception = VimSdk::VMPowerOnError.new("Could not power on VM '#{vm}': #{result.not_attempted.map(&:fault).map(&:msg).join(', ')}", result.not_attempted.first.fault)
```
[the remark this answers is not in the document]
EleanorRigby: Fixed

src/vsphere_cpi/spec/integration/drs_rules_spec.rb
ktchen14: If it doesn't work then just use the block form as before. Either way the logic should belong in `VMPoserOnError` itself. But according to the RSpec documentation it should work as described here:

https://relishapp.com/rspec/rspec-expectations/docs/composing-matchers

With the predicate matchers as described here:

https://relishapp.com/rspec/rspec-expectations/v/3-7/docs/built-in-matchers/predicate-matchers

EleanorRigby: When I try predicate and compose matcher as you described it gives me error like 
```
 expected VSphereCloud::VMPowerOnError with be unplaceable, got #<VSphereCloud::VMPowerOnError: Could not power on VM '<[Vim.VirtualMachine] vm-109>': This operation would violate a virtual machine affinity/anti-affinity rule.> with backtrace:
```
It is taking be_unplaceable as a string. 
I read the links you posted and it should work according to them. The raise_error composing matcher example is like below 
```
RSpec.describe "Passing matchers to raise_error" do
  specify "you can pass a matcher in place of the message" do
    expect {
      raise RuntimeError, "this goes boom"
    }.to raise_error(RuntimeError, a_string_ending_with("boom"))
  end
end
```
Are we understanding something wrong here?

PR #83, opened 2018-04-09
Add support to add vm to load balancer (NSXT-2.1)

nehagjain15:
DO NOT MERGE, WIP

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -313,7 +313,8 @@ module VSphereCloud
             @nsxt_provider.set_vif_type(created_vm, vm_type.nsxt)
           end
         rescue => e
-          @logger.info("Failed to apply NSXT properties to VM '#{created_vm.cid}' with error: #{e}")
+          error_message = e.is_a?(NSXT::ApiCallError) ? e.response_body : e.message
+          @logger.info("Failed to apply NSX properties to VM '#{created_vm.cid}' with error: #{error_message}")
           begin
             @logger.info("Deleting VM '#{created_vm.cid}'...")
             delete_vm(created_vm.cid)
@@ -387,10 +388,9 @@ module VSphereCloud
           end
           begin
             @nsxt_provider.remove_vm_from_server_pools(vm_ip)
-          rescue NSXT::ApiCallError => e
-            @logger.info("Failed to remove VM from ServerPool: #{e.response_body}")
           rescue => e
-            @logger.info("Failed to remove VM from ServerPool: #{e.message}")
+            error_message = e.is_a?(NSXT::ApiCallError) ? e.response_body : e.message
+            @logger.info("Failed to remove VM from ServerPool: #{error_message}")
           end
         end
 
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -301,9 +301,10 @@ module VSphereCloud
           if @config.nsxt_enabled?
             @nsxt_provider.add_vm_to_nsgroups(created_vm, vm_type.nsxt)
             @nsxt_provider.set_vif_type(created_vm, vm_type.nsxt)
+            @nsxt_provider.add_vm_to_lbs(created_vm, vm_type.nsxt_lbs) if vm_type.nsxt_lbs
           end
         rescue => e
-          @logger.info("Failed to add VM '#{created_vm.cid}' to NSGroups with error: #{e}")
+          @logger.info("Failed to apply NSXT properties to VM '#{created_vm.cid}' with error: #{e}")
           begin
             @logger.info("Deleting VM '#{created_vm.cid}'...")
             delete_vm(created_vm.cid)
@@ -314,7 +315,6 @@ module VSphereCloud
         end
 
         begin
-
           vm_type.nsx_security_groups.each do |security_group|
             nsx.add_vm_to_security_group(security_group, created_vm.mob_id)
           end unless vm_type.nsx_security_groups.nil?
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -367,8 +367,8 @@ module VSphereCloud
     def delete_vm(vm_cid)
       with_thread_name("delete_vm(#{vm_cid})") do
         @logger.info("Deleting vm: #{vm_cid}")
-
         vm = vm_provider.find(vm_cid)
+        vm_ip = vm.mob.guest&.ip_address
         vm.power_off
 
         persistent_disks = vm.persistent_disks
@@ -384,6 +384,13 @@ module VSphereCloud
             @nsxt_provider.remove_vm_from_nsgroups(vm)
           rescue => e
             @logger.info("Failed to remove VM from NSGroups: #{e.message}")
+          end
+          begin
+            @nsxt_provider.remove_vm_from_server_pools(vm_ip)
+          rescue NSXT::ApiCallError => e
+            @logger.info("Failed to remove VM from ServerPool: #{e.response_body}")
+          rescue => e
+            @logger.info("Failed to remove VM from ServerPool: #{e.message}")
           end
         end
 
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -305,6 +305,7 @@ module VSphereCloud
               #For dynamic server pools add vm to the corresponding nsgroup
               static_server_pools, dynamic_server_pools = @nsxt_provider.retrieve_server_pools(vm_type.nsxt_server_pools)
               lb_ns_groups = dynamic_server_pools.map{ |server_pool| server_pool.member_group.grouping_object.target_display_name } if dynamic_server_pools
+              @logger.info("NSGroup names corresponding to load balancer's dynamic server pools are: #{lb_ns_groups}")
               ns_groups.concat(lb_ns_groups) if lb_ns_groups
               @nsxt_provider.add_vm_to_server_pools(created_vm, static_server_pools) if static_server_pools
             end
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -299,9 +299,17 @@ module VSphereCloud
 
         begin
           if @config.nsxt_enabled?
-            @nsxt_provider.add_vm_to_nsgroups(created_vm, vm_type.ns_groups)
+            ns_groups = vm_type.ns_groups || []
+            if vm_type.nsxt_server_pools
+              #For static server pools add vm as server pool member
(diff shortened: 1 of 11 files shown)


src/vsphere_cpi/lib/cloud/vsphere/vm_type.rb
```diff
@@ -27,5 +27,13 @@ def nsx_security_groups
     def nsx_lbs
       nsx['lbs'] if nsx
     end
+
+    def nsxt_server_pools
+      nsxt['lb']['server_pools'] if nsxt && nsxt['lb']
```
ktchen14:
```ruby
nsxt&.dig('lb', 'server_pools')
```

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -198,6 +200,70 @@
         end
       end
     end
+
+    context 'when load balancers are specified' do
```
ktchen14: Did you mean when server pools are specified?

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -198,6 +200,70 @@
...
+                    'name' => server_pool_name_2,
+                    'port' => 80
+                  }
+                ]
+              }
+            }
+          }
+      end
+
+      context 'but atleast one server pool does not exist' do
```
ktchen14: Typo

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -198,6 +200,70 @@
...
+        it 'raises an error' do
+          expect do
+            simple_vm_lifecycle(cpi, @nsxt_opaque_vlan_1, vm_type)
+          end.to raise_error(VSphereCloud::ServerPoolsNotFound)
+        end
+      end
+      context 'and all server pool exists' do
```
ktchen14: And all server pools exist

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -306,6 +377,29 @@ def nsgroup_effective_logical_port_member_ids(nsgroup)
     results.map { |member| member.target_id }
   end
 
+  def create_server_pool(pool_name, nsgroup=nil)
```
ktchen14: I'd rather we have two methods to create the two kinds of server pools. Also spaces around `=`.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -276,11 +276,21 @@ def create_vm(agent_id, stemcell_cid, vm_type, networks_spec, existing_disk_cids
 
         begin
           if @config.nsxt_enabled?
-            @nsxt_provider.add_vm_to_nsgroups(created_vm, vm_type.nsxt)
+            ns_groups = vm_type.ns_groups || []
```
ktchen14: This logic is complicated enough that I'd rather move this stuff into a separate module and have a stack of modules like middleware.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -197,6 +220,51 @@ def set_vif_type(vm, vm_type_nsxt)
       end
     end
 
+    # Add vm to given list of static Load Balancer Server Pools
+    # @param [Resources::VM] vm
+    # @param [[NSXT::LbPool, integer][]] load_balancer_pools Array of Load_balancer_pool and port no
+    def add_vm_to_server_pools(vm, load_balancer_pools)
+      return if load_balancer_pools.nil? ||  load_balancer_pools.empty?
```
ktchen14: Some weird extra spaces here and down below.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -197,6 +220,51 @@ def set_vif_type(vm, vm_type_nsxt)
...
+        end
+      end
+    end
+
+    # Returns an array of Static Load Balancer Server Pools and Dynamic Load Balancer Pools
+    # For Static Server Pools corresponding port no is also returned
+    # @param [array] server_pools It is an array of hashes with server_pool names and port
+    def retrieve_server_pools(server_pools)
+      return [] if server_pools.nil? || server_pools.empty?
+      server_pools_by_name = services_svc.list_load_balancer_pools.results.each_with_object({}) do |server_pool, hash|
```
ktchen14: This will cause a problem if two server pools have the same display name.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -197,6 +220,51 @@ def set_vif_type(vm, vm_type_nsxt)
...
+        hash[server_pool.display_name] = server_pool
+      end
+
+      missing = server_pools.reject do |server_pool|
+        server_pools_by_name.key?(server_pool['name'])
+      end
+      raise ServerPoolsNotFound.new(*missing) unless missing.empty?
+
+      static_server_pools, dynamic_server_pools = [],[]
```
ktchen14: Use the `Enumerable#partition` method for this logic.

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -198,6 +200,70 @@
...
+          end.to raise_error(VSphereCloud::ServerPoolsNotFound)
+        end
+      end
+      context 'and all server pool exists' do
+        let(:nsgroup_1) { create_nsgroup(nsgroup_name_1) }
+        let(:server_pool_1) { create_server_pool(server_pool_name_1) }
+        let(:server_pool_2) { create_server_pool(server_pool_name_2, nsgroup_1) }
+
+        before do
+          expect(server_pool_1).to_not be_nil
```
ktchen14: This logic belongs in `create_server_pool`. You can force a `let` to be strict rather than lazy with `let!`.

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -306,6 +377,29 @@ def nsgroup_effective_logical_port_member_ids(nsgroup)
...
+  def create_server_pool(pool_name, nsgroup=nil)
+    server_pool = NSXT::LbPool.new(display_name: pool_name)
+    if nsgroup
+      resource = NSXT::ResourceReference.new(target_id: nsgroup.id, target_type: nsgroup.resource_type, target_display_name: nsgroup.display_name)
+      server_pool.member_group = NSXT::PoolMemberGroup.new(grouping_object: resource, max_ip_list_size: 2)
+    end
+    services_svc.create_load_balancer_pool(server_pool)
+  end
+
+  def verify_pool_member(server_pool, ip_address, port_no)
```
ktchen14: Can we get rid of this function and just do something like:

```ruby
expect(...).to include(...)
```

Putting assertions in a function makes them a lot more opaque when reading the test.

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -306,6 +377,29 @@ def nsgroup_effective_logical_port_member_ids(nsgroup)
...
+    server_pool = services_svc.read_load_balancer_pool(server_pool.id)
+    matching_members = server_pool.members.select{ |member| member.ip_address == ip_address && member.port == port_no }
```
ktchen14: This line is way too long for `{ ... }`. Use `do ... end`.

src/vsphere_cpi/lib/cloud/vsphere/vm_type.rb
```diff
@@ -27,5 +27,13 @@ def nsx_security_groups
...
+    end
+
+    def ns_groups
```
ktchen14: Did we change this to `ns_groups` from `nsgroups`?

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -198,6 +200,70 @@
...
+        after do
+          delete_server_pool(server_pool_1)
+          delete_server_pool(server_pool_2)
+          delete_nsgroup(nsgroup_1)
+        end
+
+        it 'adds vm to existing static server pools and adds all logical ports of the VM to NSGroups associated with the dynamic server pool' do
+          simple_vm_lifecycle(cpi, @nsxt_opaque_vlan_1, vm_type) do |vm_id|
+            vm = cpi.vm_provider.find(vm_id)
+            expect(vm).to_not be_nil
```
ktchen14: These aren't actually necessary in a test. If they're null we'll just get exceptions thrown which will also fail the example.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -469,4 +468,94 @@
...
+    before do
+      allow_any_instance_of(VSphereCloud::NSXTProvider).to receive(:services_svc).and_return(services_svc)
+    end
+    it 'does nothing when there are no pools' do
+      nsxt_provider.retrieve_server_pools([])
+      nsxt_provider.retrieve_server_pools(nil)
+    end
+    context 'when server_pools are present' do
+      let(:serverpool_1) do
+        NSXT::LbPool.new(:id => 'id-1', :display_name => 'test-static-serverpool-1')
```
ktchen14: Use the new hash syntax for symbol keys.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -469,4 +468,94 @@
...
+  describe 'retrieve_server_pools' do
+    before do
+      allow_any_instance_of(VSphereCloud::NSXTProvider).to receive(:services_svc).and_return(services_svc)
+    end
+    it 'does nothing when there are no pools' do
+      nsxt_provider.retrieve_server_pools([])
+      nsxt_provider.retrieve_server_pools(nil)
+    end
+    context 'when server_pools are present' do
+      let(:serverpool_1) do
```
ktchen14: This should be `:server_pool_1` to keep in line with the other code.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
nehagjain15: you are right, I will chat with Yash tomorrow about this. I think the right approach might be to add it to all server pools matching the given name. I will update the test/logic accordingly

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
nehagjain15: Cannot as I have updated the logic and actually not splitting array into 2 as both arrays have different elements. Static pools also have ports being returned with them.

src/vsphere_cpi/lib/cloud/vsphere/vm_type.rb
nehagjain15: in cloud config it is actually ns_group. One of the tests was not implemented right where the config was incorrect. Also, we reference it as ns_group and nsgroup at different places. I used what we had as per the config. Need to refactor overall to match..

PR #97, opened 2018-05-07
Add global upgrade hardware option

EleanorRigby:
- Add unit test
- Add capability to override global config if specified in vm_type
block.


jobs/vsphere_cpi/spec
```diff
@@ -26,6 +26,8 @@ properties:
     description: Datacenters in vCenter to use (value is an array of Hashes representing datacenters and clusters, See director.yml.erb.erb)
   vcenter.enable_auto_anti_affinity_drs_rules:
     description: Creates DRS rule to place VMs on separate hosts.
+  vcenter.upgrade_hw_version:
+    description: Upgrades virtual machines to latest virtual harware version supported on the esx host.
```
nehagjain15: typos - Upgrades virtual machines to latest virtual hadrware version supported on the ESXi host.

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_config_spec.rb
```diff
@@ -20,6 +20,41 @@ def fake_datastore(name, free_space, mob = nil)
     let(:cluster_picker) { ClusterPicker.new(0, 0) }
     let(:cluster_provider) { nil }
 
+    describe '#upgrade_hw_version?' do
+      context 'when upgrade hardware version is nil (not specified) in both vm_type and global config' do
+        let(:cloud_properties) {{}}
+        let(:vm_type) { VmType.new(datacenter, cloud_properties)}
```
nehagjain15: definition of vm_type is same as what is in the outer block so you can remove it from all 3 context

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_config_spec.rb
```diff
@@ -20,6 +20,41 @@ def fake_datastore(name, free_space, mob = nil)
...
+        let(:input) { { vm_type: vm_type } }
```
nehagjain15: input is same for all 3 context, so it can be defined under describe '#upgrade_hw_version?' block

PR #100, opened 2018-05-10
Feature/create network spike

nehagjain15:


src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -662,6 +662,50 @@ def info
       {'stemcell_formats' =>  ['vsphere-ovf', 'vsphere-ova']}
     end
 
+    #creates T1 router and virtual switch attached to it
+    def create_subnet(subnet_definition)
+      raise 'NSXT must be enabled in CPI to use create_subnet' if !@config.nsxt_enabled?
+      cloud_properties = subnet_definition['cloud_properties']
+      raise 'cloud_properties must be provided' if cloud_properties.nil?
+      raise 'edge_cluster_id cloud property can not be empty' if cloud_properties['edge_cluster_id'].nil?
```
ktchen14: The error message doesn't match the test. Either:

```ruby
raise 'edge_cluster_id cloud property must be specified' unless cloud_properties.key?('edge_cluster_id')
```

Or:

```ruby
raise 'edge_cluster_id cloud property can not be empty' if cloud_properties['edge_cluster_id']&.empty?
```

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -662,6 +662,50 @@ def info
       {'stemcell_formats' =>  ['vsphere-ovf', 'vsphere-ova']}
     end
 
+    #creates T1 router and virtual switch attached to it
+    def create_subnet(subnet_definition)
+      raise 'NSXT must be enabled in CPI to use create_subnet' if !@config.nsxt_enabled?
+      cloud_properties = subnet_definition['cloud_properties']
+      raise 'cloud_properties must be provided' if cloud_properties.nil?
```
ktchen14: Input validation should be moved into its own class. We're moving away from the use of passing schemed objects as hashes. See `VmConfig` or `VmType`.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -767,5 +811,16 @@ def disk_configurations(vm_type, existing_disk_cids=[])
       )
       disk_configurations.push(ephemeral_disk_config)
     end
+
+    #This subnet will be used in create_logical_router_port.
+    # ip_addresses is going to be a gateway IP for subnet - aka IP for router.
+    def create_subnet_obj(range, gateway)
+      if (!range.nil? && range.include?('/'))
+        _, mask = range.split("/")
+        return NSXT::IPSubnet.new({:ip_addresses => [gateway],
```
ktchen14: Use new style Ruby hashes:

```ruby
return NSXT::IPSubnet.new(ip_addresses: [gateway], prefix_length: mask.to_i)
```

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -767,5 +811,16 @@ def disk_configurations(vm_type, existing_disk_cids=[])
       )
       disk_configurations.push(ephemeral_disk_config)
     end
+
+    #This subnet will be used in create_logical_router_port.
+    # ip_addresses is going to be a gateway IP for subnet - aka IP for router.
+    def create_subnet_obj(range, gateway)
+      if (!range.nil? && range.include?('/'))
```
ktchen14: Use a proper regular expression here. This won't be caught by this logic:

```
192.168.1.1/24/24
```

This probably works: https://www.regextester.com/93987

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -287,6 +288,108 @@ def remove_vm_from_server_pools(vm_ip)
       end
     end
 
+    def create_t1_router(edge_cluster_id, name = nil)
+      name = SecureRandom.base64(8) if name.nil?
+      router_api.create_logical_router({ :edge_cluster_id => edge_cluster_id,
```
ktchen14: Use new style Ruby hashes.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -662,6 +662,50 @@ def info
...
+      switch_id = nil
+      begin
+        t1_router = @nsxt_provider.create_t1_router(cloud_properties['edge_cluster_id'], cloud_properties['t1_name'])
+        t1_router_id = t1_router.id
+        @nsxt_provider.enable_route_advertisement(t1_router_id)
+        @nsxt_provider.attach_t1_to_t0(cloud_properties['t0_router_id'], t1_router_id)
+        switch = @nsxt_provider.create_logical_switch(cloud_properties['transport_zone_id'], cloud_properties['switch_name'])
+        switch_id = switch.id
+        @nsxt_provider.attach_switch_to_t1(switch_id, t1_router_id, subnet)
+      rescue Exception => e
```
ktchen14: Rescuing `Exception` is almost never what you want. We should be specific about the possible error types here.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -662,6 +662,50 @@ def info
...
+        @logger.error('Failed to create subnet. Trying to clean up')
+        @nsxt_provider.delete_t1_router(t1_router_id) unless t1_router_id.nil?
+        @nsxt_provider.delete_logical_switch(switch_id) unless switch_id.nil?
+
+        raise "Failed to create subnet. Has router been created: #{!t1_router_id.nil?}. Has switch been created: #{!switch_id.nil?}. Exception: #{e}"
+      end
+      {:network_cid => switch.id, :cloud_properties => {:name => switch.display_name } }
```
ktchen14: This return type needs to be a real type and not a hash. For one thing the fact that `cloud_properties` on input is a string key and on output is a symbol key is confusing.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -662,6 +662,50 @@ def info
...
+    def create_subnet(subnet_definition)
+      raise 'NSXT must be enabled in CPI to use create_subnet' if !@config.nsxt_enabled?
+      cloud_properties = subnet_definition['cloud_properties']
+      raise 'cloud_properties must be provided' if cloud_properties.nil?
+      raise 'edge_cluster_id cloud property can not be empty' if cloud_properties['edge_cluster_id'].nil?
+      raise 't0_router_id cloud property can not be empty' if cloud_properties['t0_router_id'].nil?
+      raise 'transport_zone_id cloud property can not be empty' if cloud_properties['transport_zone_id'].nil?
+      subnet = create_subnet_obj(subnet_definition['range'], subnet_definition['gateway'])
+
+      t1_router_id = nil
```
ktchen14: These two variables aren't really necessary. It looks like they're here to facilitate the `rescue` clause. Either put each `@nsxt_provider` action into its own `begin ... rescue ... end` block or use something like:

```ruby
@nsxt_provider.delete_t1_router(t1_router.id) if defined? t1_router
```

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -767,5 +811,16 @@ def disk_configurations(vm_type, existing_disk_cids=[])
...
       disk_configurations.push(ephemeral_disk_config)
     end
+
+    #This subnet will be used in create_logical_router_port.
+    # ip_addresses is going to be a gateway IP for subnet - aka IP for router.
+    def create_subnet_obj(range, gateway)
+      if (!range.nil? && range.include?('/'))
+        _, mask = range.split("/")
+        return NSXT::IPSubnet.new({:ip_addresses => [gateway],
+                                   :prefix_length => mask.to_i})
```
ktchen14: Ruby has multiple modes of typecasting that are well explained here:

https://idiosyncratic-ruby.com/54-try-converting.html

In short though `#to_i` is almost never what you want to use with user input since:

> This method never raises an exception when base is valid

Instead you probably want something like `Integer(mask, 10)`.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
medvedzver: We need to clean up what we created. Also exception thrown by nsxt client are empty. So I think it's nice to have a meaningful message.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -662,6 +662,50 @@ def info
...
+        raise "Failed to create subnet. Has router been created: #{!t1_router_id.nil?}. Has switch been created: #{!switch_id.nil?}. Exception: #{e}"
+      end
+      {:network_cid => switch.id, :cloud_properties => {:name => switch.display_name } }
+    end
+
+    def delete_subnet(switch_id)
+      raise 'NSXT must be enabled in CPI to use delete_subnet' if !@config.nsxt_enabled?
+      raise 'switch id must be provided for deleting a subnet' if switch_id.nil?
+      t1_router_ids = @nsxt_provider.get_attached_router_ids(switch_id)
+      raise "Expected switch #{switch_id} to have one router attached. Found #{t1_router_ids.length}" if t1_router_ids.length != 1
```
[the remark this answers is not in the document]
medvedzver: We had a meeting were we discussed this use case. We decided to do the same thing as PKS does if someone messing up with infrastructure - fail with error. To avoid deleting users's stuff.

jobs/vsphere_cpi/spec
```diff
@@ -57,6 +57,9 @@ properties:
       -----BEGIN CERTIFICATE-----
       MII...
       -----END CERTIFICATE-----
+  vcenter.nsxt.skip_ssl_verify:
```
cppforlife: dont want to add skipping ssl verification as a feature.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -662,6 +662,50 @@ def info
       {'stemcell_formats' =>  ['vsphere-ovf', 'vsphere-ova']}
     end
 
+    #creates T1 router and virtual switch attached to it
+    def create_subnet(subnet_definition)
```
cppforlife: methods should be called create_network and delete_network

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -287,6 +288,108 @@ def remove_vm_from_server_pools(vm_ip)
...
+                                         :router_type => 'TIER1',
+                                         :display_name => name})
+    end
+
+    def attach_t1_to_t0(t0_router_id, t1_router_id)
+      begin
+        t0_router_port =  NSXT::LogicalRouterLinkPortOnTIER0.new({:logical_router_id => t0_router_id,
+                                                                  :resource_type => 'LogicalRouterLinkPortOnTIER0'})
+        t0_router_port = router_api.create_logical_router_port(t0_router_port)
+      rescue Exception => e
```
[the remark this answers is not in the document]
medvedzver: The reason for rescue here is that `attach_t1_to_t0` doing 2 API calls to NSXT. 
If one of them fails - you have to go to source code and find what line caused the issue. I think this  will help you to identify error faster.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
cppforlife: probably should use netaddr gem (part of standard library) to parse the range.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -287,6 +288,108 @@ def remove_vm_from_server_pools(vm_ip)
...
+        t0_reference = NSXT::ResourceReference.new({:target_id => t0_router_port.id,
+                                                    :target_type => 'LogicalRouterLinkPortOnTIER0',
+                                                    :is_valid => true })
+        t1_router_port = NSXT::LogicalRouterLinkPortOnTIER1.new({:linked_logical_router_port_id => t0_reference,
+                                                                 :logical_router_id => t1_router_id,
+                                                                 :resource_type => 'LogicalRouterLinkPortOnTIER1'})
+        router_api.create_logical_router_port(t1_router_port)
+      rescue Exception => e
+        @logger.error("Error creating port on T1 router #{t1_router_id} and attaching it to T0 port #{t0_router_port.id}. Exception: #{e}")
+        raise "Error creating port on T1 (#{t1_router_id}) and attaching it to T0 port #{t0_router_port.id}. Exception: #{e}"
```
cppforlife: use `e.inspect` instead of just `e` so that its clear which class name error is from since a lot of exception in ruby dont have *unique* descriptions.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
cppforlife: scope in ruby i believe is shared so you dont need to pull out there vars above, they will be defined from within begin and be available in rescue.

> if defined? t1_router

that wouldnt work, since defined? return "local-variable" always even if it's nil.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
medvedzver: Was thinking about it. I understand that we are creating network infrastructure, but on director side you asking to create a subnet and you give a subnet definition. 
As  I remember your point for calling this `*_network` was that we are passing `networks_spec` to VM. 
Would it be more confusing if we call `create_network` on a BOSH side and give it a `subnet_definition` and it returns us `cloud_properties` for a subnet?

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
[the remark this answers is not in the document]
ktchen14: No problem that’s fine but rescuing Exception is very dangerous in Ruby. For example here:

https://stackoverflow.com/questions/10048173/why-is-it-bad-style-to-rescue-exception-e-in-ruby

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
ktchen14: @cppforlife On my end it looks like netaddr isn’t part of the standard library which is why I recommended the Regexp. Did it get added recently?

medvedzver: Going with `netaddr`

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
medvedzver: @cppforlife ^^

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
ktchen14: You’re right. Looks like the check should be t1_router.nil?

src/vsphere_cpi/lib/cloud/vsphere/network.rb
```diff
@@ -0,0 +1,189 @@
...
+      @switch_provider = switch_provider
+      @router_provider = router_provider
+      @ip_block_provider = ip_block_provider
+    end
+
+    #creates T1 router and virtual switch attached to it
+    def create(network_definition)
+      validate(network_definition)
+      begin
+        cloud_properties = network_definition['cloud_properties']
```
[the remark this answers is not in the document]
medvedzver: @ktchen14 I think we also need to rename network to NetworkProvider as we discussed.
As for `active record` approach - I think we don't need to introduce it yet. Right now we are using `Provider` model in some other places (like VmCreator). I would just stick with it.
What do you think?

ktchen14:
Merged in dfe878579cbab768af07a12bb5543cd016cbb762.

PR #102, opened 2018-05-14
Change call to delete_vm to cleanup properly

EleanorRigby:
- In vm_creator.rb, occasionally vsphere or CPI might fail to power on
the VM OR set the network properties OR set the agent metadata OR fail
while performing some other vm related ops. In these failed scenarios,CPI
deletes the cloned vm before bubbling up the error to the logs.
While deleting the vm, it must also clean additional agent files & disks that
were attached to vm. This change ensures that.
Old call of created_vm.delete directly deletes a vm without cleaning vm's agent
data and detaching disk. Replacing call with cpi.delete_vm(created_vm.cid)
makes sure to clean things up before deleting the vm.

[Delivers #156900205](https://www.pivotaltracker.com/story/show/156900205)


src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -142,7 +142,7 @@ def create(vm_config)
         e.vm_cid = vm_config.name if e.instance_of?(Cloud::NetworkException)
         @logger.info("#{e} - #{e.backtrace.join("\n")}")
         begin
-          created_vm.delete if created_vm
+          cpi.delete_vm(created_vm.cid) if created_vm
```
nehagjain15: should be @cpi.delete_vm

PR #103, opened 2018-05-15
Make logger a singleton

ktchen14:
Still a work in progress. After doing:

```ruby
require 'cloud/vsphere/logger'
```

It's possible to do:

```ruby
extend VSphereCloud::Logger
```

To gain `logger` as both an instance and a singleton method. Basically every file has been touched but the actual change is present in `src/vsphere_cpi/lib/cloud/vsphere/logger.rb`.

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -26,7 +26,7 @@ module VSphereCloud
         vcenter_api_uri: @config.vcenter_api_uri,
         http_client: http_client
       )
-      def other_client.logger
+      other_client.define_singleton_method(:logger) do
         @logger ||= ::Logger.new('/dev/null')
       end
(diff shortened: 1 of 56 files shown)


src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -15,16 +17,18 @@ def message
...
 
     def enable_telemetry
       http_client = VSphereCloud::CpiHttpClient.new
       other_client = VCenterClient.new(
         vcenter_api_uri: @config.vcenter_api_uri,
-        http_client: http_client,
-        logger: Logger.new('/dev/null'),
+        http_client: http_client
       )
+      def other_client.logger
```
nehagjain15: Seeing def inside def doesnt look nice. Could we instead do other_client.define_sigleton_method(:logger) ?

cppforlife:
-1. this effectively introduces global variable `logger`, and global vars arent ideal. 

to give a concrete side effect, one can no longer run unit tests safely in parallel because now setting logger is racy.

PR #106, opened 2018-05-30
Refactor create disk/attach disk/create vm for selection pipeline design

EleanorRigby:
This is implementation of create disk selection pipeline


src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
```diff
@@ -0,0 +1,85 @@
...
+      @filter_list ||= []
+    end
+
+    def scorer_list
+      @scorer_list ||= []
+    end
+  end
+
+  class DiskPlacementSelectionPipeline < SelectionPipeline
+
```
EleanorRigby: Remove this blank line

EleanorRigby:
This is almost complete and this is ready for review

EleanorRigby:
Ready for review.

src/vsphere_cpi/lib/cloud/vsphere/filters.rb
```diff
@@ -0,0 +1,30 @@
...
+    def call(storage_placement, criteria_object)
+      storage_placement.maintenance_mode == "normal"
+    end
+  end
+
+  module Filter_Free_Space_DS
+    module_function
+    DEFAULT_DISK_HEADROOM = 1024
+    def call(storage_placement, criteria_object)
+      storage_placement.free_space > DEFAULT_DISK_HEADROOM + criteria_object[:size_in_mb]
```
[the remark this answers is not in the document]
EleanorRigby: I was thinking that criteria object will be different for each pipeline and different types of placement might need differing criteria. Will it be good to just keep it as a hash like now?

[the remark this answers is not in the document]
EleanorRigby: cool.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -497,33 +497,52 @@ def detach_disk(vm_cid, raw_director_disk_cid)
...
+          accessible_datastores.values.collect do |ds|
+            StoragePlacement.new(ds)
+          end
+        end
 
-        disk_type = cloud_properties.fetch('type', @config.vcenter_default_disk_type)
-        disk = @datacenter.create_disk(datastore, size_in_mb, disk_type)
-        logger.info("Created disk: #{disk.inspect}")
+        # For each possible placement try to create a disk
+        disk_select_pipe.each do |storage_placement|
```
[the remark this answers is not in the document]
EleanorRigby: I am returning a lazy enumerator if block is not given. Not sure how to break this into series of blocks.
Do you mean something like this 
```
        placement_enumerator = pipeline.each
        placement_enumerator.each do |storage_placement|
          if vm_cid
            # This is to take into account datastore which might be accessible from cluster defined in cloud config
            # #datacenter.accessible_datastores includes datastores based on global config, so cannot use that here
            vm = vm_provider.find(vm_cid)
            datastore = vm.accessible_datastores[storage_placement.name]
            raise "Can't find datastore '#{storage_placement.name}'" if datastore.nil?
          else
            datastore = @datacenter.find_datastore(storage_placement.name)
          end
          datastore
        end.each do |ds|
          logger.info("Trying disk creation : Using datastore #{datastore.name} to store persistent disk")
          disk_type = cloud_properties.fetch('type', @config.vcenter_default_disk_type)
          disk = @datacenter.create_disk(datastore, size_in_mb, disk_type)
          next if disk.nil?
          logger.info("Created disk: #{disk.inspect}")
          raw_director_disk_cid = disk_pool.storage_list.any? ? DirectorDiskCID.encode(disk.cid, target_datastore_pattern: target_datastore_pattern) : disk.cid
          # Return disk cid if we create a valid disk
          return raw_director_disk_cid unless disk.nil?
        end
```

src/vsphere_cpi/lib/cloud/vsphere/filters.rb
```diff
@@ -0,0 +1,30 @@
+module VSphereCloud
```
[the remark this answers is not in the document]
EleanorRigby: Agree

src/vsphere_cpi/lib/cloud/vsphere/placement.rb
```diff
@@ -0,0 +1,66 @@
...
+    attr_reader :resource, :score
+
+    METHOD_NOT_IMPLEMENTED = "method is not overridden/implemented in sub class"
+
+    def eql?(other)
+      other.resource == resource && other.score == score
+    end
+
+    def hash
+      [resource].hash
```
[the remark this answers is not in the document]
EleanorRigby: Then we will have to define hash method in each resource right?

src/vsphere_cpi/lib/cloud/vsphere/filters.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  module Filter_Free_Space_DS
+    module_function
+    DEFAULT_DISK_HEADROOM = 1024
+    def call(storage_placement, criteria_object)
+      storage_placement.free_space > DEFAULT_DISK_HEADROOM + criteria_object[:size_in_mb]
+    end
+  end
+
+  module Filter_Target_Pattern_DS
```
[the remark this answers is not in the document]
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/placement.rb
```diff
@@ -0,0 +1,66 @@
+module VSphereCloud
+  class Placement
+    include Comparable
+
+    attr_reader :resource, :score
+
+    METHOD_NOT_IMPLEMENTED = "method is not overridden/implemented in sub class"
+
+    def eql?(other)
```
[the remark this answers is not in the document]
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/placement.rb
```diff
@@ -0,0 +1,66 @@
...
+    def inspect
+      raise NotImplementedError, "#{METHOD_NOT_IMPLEMENTED}"
+    end
+
+    def initialize(resource)
+      @resource = resource
+      @score = 0
+    end
+
+    def <=>(other)
```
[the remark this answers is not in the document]
EleanorRigby: `1 <=> b # 0` Not sure how did that work. Should have thrown Argument Error.
`a==b # false` this is right as they do not have same score. 

I am not able to follow as to what  kind of improvement you are thinking here.

src/vsphere_cpi/lib/cloud/vsphere/placement.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def eql?(other)
+      other.resource == resource && other.score == score
+    end
+
+    def hash
+      [resource].hash
+    end
+
+    def inspect
```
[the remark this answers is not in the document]
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/placement.rb
```diff
@@ -0,0 +1,66 @@
...
+      resource.free_space
+    end
+
+    # @return [Boolean] accessibility of the resource
+    def accessible?
+      resource.accessible?
+    end
+
+    # @return [Boolean] whether resource is in normal maintenance mode state
+    def maintenance_mode
```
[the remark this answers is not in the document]
EleanorRigby: Not actually a boolean. it is a string representing the mode the resource is in. I will change the comments.

src/vsphere_cpi/lib/cloud/vsphere/placement.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    # @return [String] name of the resource
+    def name
+      resource.name
+    end
+
+    # @return [String] debug StoragePlacement information.
+    def inspect
+      "<Placement has resource : #{resource.inspect} and score #{score}>"
```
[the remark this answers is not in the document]
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
```diff
@@ -0,0 +1,85 @@
+module VSphereCloud
+  class SelectionPipeline
+    include Enumerable
+
+    # @return [Object] the object to select placements for
+    attr_reader :object
+
+    # @param object [Object] the object to select placements for
+    def initialize(object, *args)
+      raise ArgumentError, 'No gather block provided' unless block_given?
```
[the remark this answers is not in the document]
EleanorRigby: I want to be able to enforce design and developer constraints by raising an error so that any new pipeline developer needs to provide a block. 
What advantage does an empty list offer? We are not allowing for addition of gather block later.

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    def accept?(placement)
+      filter_list.all? do |filter|
+        filter.call(placement, @object)
+      end
+    end
+
+    def score(placement)
+      scorer_list.each do |scorer|
```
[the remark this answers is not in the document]
EleanorRigby: Changed it to reduce. If we take @score out of placement , how are we supposed to sort or compare them

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    def with_scorer(*args)
+      if block_given?
+        unless args.empty?
+          raise ArgumentError, "wrong number of arguments (given #{args.length}, expected 0)"
+        end
+        scorer_list << Proc.new
+      else
+        scorer_list.concat(args)
```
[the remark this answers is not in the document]
EleanorRigby: Concat works with empty arguments too. I can raise error like `expected at least one arg`

src/vsphere_cpi/lib/cloud/vsphere/resources/storage_pod.rb
```diff
@@ -42,6 +56,12 @@ def self.find_storage_pod(name, datacenter_mob)
       def datastores
         mob.child_entity
       end
+
+      # @return [String] debug storagePod information.
+      def inspect
+        "<StoragePod: #@mob / #{name}>"
```
nehagjain15: Did you want to print the mob here? #{mob} ?

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -506,38 +512,46 @@ def detach_disk(vm_cid, raw_director_disk_cid)
     def create_disk(size_in_mb, cloud_properties, vm_cid = nil)
       with_thread_name("create_disk(#{size_in_mb}, _)") do
         logger.info("Creating disk with size: #{size_in_mb}")
-        if vm_cid
-          vm = vm_provider.find(vm_cid)
-          accessible_datastores = vm.accessible_datastores
-        else
-          accessible_datastores = @datacenter.accessible_datastores
-        end
+        # Create a disk pool to hold possible datastyores
```
nehagjain15: typo in datastores spelling

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -18,154 +18,170 @@ def initialize(client:, cloud_searcher:, cpi:, datacenter:, agent_env:, ip_confl
     end
 
     def create(vm_config)
-      cluster = vm_config.cluster
-      storage = StoragePicker.choose_ephemeral_storage(vm_config.ephemeral_datastore_name, vm_config.cluster.accessible_datastores, vm_config.vm_type)
+      created_vm = nil
+      vm_config.cluster_placements.each do |cluster_placement|
+        # Get the  cluster from the placement
+        cluster = cluster_placement.cluster
```
EleanorRigby: change the name to vm placement

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -18,154 +18,170 @@ def initialize(client:, cloud_searcher:, cpi:, datacenter:, agent_env:, ip_confl
...
+        # Validates if the rule specified is of right syntax
+        vm_config.validate_drs_rules(cluster)
 
-      datastore, datastore_cluster = storage.is_a?(Resources::StoragePod) ? [nil, storage] : [storage, nil]
+        #@TA: TODO Choose host here if needed.
```
EleanorRigby: remove this or make it just todo

src/vsphere_cpi/lib/cloud/vsphere/resources/storage_pod.rb
EleanorRigby: This is the usual way every resource inspect has been implemented. So, I followed the convention. I do not know why it is done this way. 
I think it does give us a clue on the mob name and makes it easier to find in managed object browser of vsphere.

I will let it be for now

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
EleanorRigby: Fixed

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
EleanorRigby: This also I can keep for future changes. As it will lead to lot of changes in unit test. I want things to be just working unless it is a breaking bug

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_placement_selection_pipeline_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  context 'when there are no disks to be configured' do
+    it 'only generates placements with enough memory' do
+      allow_any_instance_of(VSphereCloud::Resources::Datastore).to receive(:maintenance_mode?).and_return(false)
+      allow(cluster_1).to receive(:free_memory).and_return(1024)
+      allow(cluster_2).to receive(:free_memory).and_return(2048)
+      expect(subject.to_a).to contain_exactly(placement_2)
+    end
+
+    it 'only generates placements with atleast one of the datastores not in maintenance mode' do
```
ktchen14: Typo

src/vsphere_cpi/spec/integration/datastore_cluster_spec.rb
```diff
@@ -56,7 +56,7 @@
         [],
         {}
       )
-    end.to raise_error(/No valid placement found for disks/)
+    end.to raise_error(/No valid placement /)
```
ktchen14: Extra space.

src/vsphere_cpi/spec/integration/datastore_cluster_spec.rb
```diff
@@ -187,7 +187,7 @@
         [],
         {}
       )
-      end.to raise_error(/No valid placement found for disks/)
+      end.to raise_error(/No valid placement found/)
```
ktchen14: Why're we looking for a different error message than on line 59?

src/vsphere_cpi/spec/integration/datastore_migration_spec.rb
```diff
@@ -136,8 +136,7 @@
         mismatch_cpi.create_vm('agent-007', @stemcell_id, vm_type, network_spec)
       }.to raise_error { |error|
         expect(error).to be_a(Bosh::Clouds::CloudError)
-        expect(error.message).to include(@second_cluster_datastore)
-        expect(error.message).to include('No valid placement found for disks')
+        expect(error.message).to include('No valid placement found')
```
ktchen14: Let's take this opportunity to fix this code to:

```ruby
}.to raise_error(Bosh::Cloud::CloudError, /No valid placement found/)
```

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/storage_pod_spec.rb
```diff
@@ -20,6 +20,38 @@
     end
   end
 
+  describe '#inspect' do
```
ktchen14: We don't need to test `#inspect`. This method should only be used in debugging code. If we're using it in production we're doing it wrong.

src/vsphere_cpi/spec/unit/cloud/vsphere/selection_pipeline_spec.rb
```diff
@@ -0,0 +1,170 @@
+require 'spec_helper'
+require 'cloud/vsphere/selection_pipeline'
+
+describe VSphereCloud::SelectionPipeline do
+  subject { described_class.new(criteria) { placement } }
+  let(:criteria) { :fake_criteria }
+  let(:placement) { [:fake_placement] }
+
+  class Integer
```
ktchen14: ???

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_placement_selection_pipeline_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+  def fake_cluster(name, *args)
+    instance_double('VSphereCloud::Resources::Cluster',
+      :name => name,
+      :free_memory => 8192,
+      :accessible_datastores => args.map{|ds| [ds.name, ds]}.to_h )
+  end
+
+  let(:criteria) { [disk_config: disk_config, req_memory: 1024] }
```
ktchen14: Are all of these required at the top level? It seems like there's a lot of rspec magic going on in this spec file. Can we organize it differently to make it a little easier to read?

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
```diff
@@ -0,0 +1,125 @@
...
+      else
+        raise ArgumentError, "0 arguments passed, expected at least 1 arg or a block if no args provided" if args.empty?
+        scorer_list.concat(args)
+      end
+      self
+    end
+
+    private
+
+    def compare_placements
```
ktchen14: While `#compare_placements` conforms to Ruby's `Comparable` module's semantics I really dislike this use in the selection pipeline because it's unclear which placement is considered "better" is not obvious.

I think each scorer should be passed `p1` and `p2` and return the "better" placement or `nil` if the two placements are considered equivalent.

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
```diff
@@ -0,0 +1,125 @@
...
+      lambda do |p1, p2|
+        scorer_list.each do |scorer|
+          result = scorer.call(p1, p2)
+          if result != 0
+            return result
```
nehagjain15: Also, there should be some logging for scorer without that we wont be able to say based on what score(comparison) the placement was picked

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    def inspect
+      "#<SelectionPipeline @filter_list #{filter_list} @scorer_list #{scorer_list}>"
+    end
+
+    def accept?(placement)
+      result = filter_list.all? do |filter|
+        result = filter.call(placement, @object)
+        logger.info("Rejecting #{placement.inspect}") unless result
```
nehagjain15: here as well, we should add based on what filter a placement was rejected will be more informative

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/storage_pod_spec.rb
EleanorRigby: Fixed

src/vsphere_cpi/spec/integration/datastore_migration_spec.rb
EleanorRigby: Fixed

src/vsphere_cpi/spec/unit/cloud/vsphere/selection_pipeline_spec.rb
EleanorRigby: Logging is using inspect method to print out certain messages. Inspect is not implemented for the for the fake integer placements we have. That's why I just introduced a dummy one here.

src/vsphere_cpi/spec/integration/datastore_cluster_spec.rb
EleanorRigby: I think I copied at different times. Considering that it is a regex , I did not catch it. Correcting it.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -516,38 +524,49 @@ def detach_disk(vm_cid, raw_director_disk_cid)
     def create_disk(size_in_mb, cloud_properties, vm_cid = nil)
       with_thread_name("create_disk(#{size_in_mb}, _)") do
         logger.info("Creating disk with size: #{size_in_mb}")
```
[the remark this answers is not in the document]
EleanorRigby: Fixed

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_placement_selection_pipeline_spec.rb
EleanorRigby: The idea is to simulate a two cluster scenario with two individual datastores and a shared datastore. I declared these at the top to simulate a mini datacenter for all the examples.Should I document with some comments? 

It can be made specific for each example in there. But I prefer to make a large playground(dummy-datacenter) and then write examples with allowing/disallowing properties of various entities.

What do you think?

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_placement_selection_pipeline_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+
+  def fake_datastore(name, free_space: 4096)
+    VSphereCloud::Resources::Datastore.new(
+      name, true, instance_double('VimSdk::Vim::Datastore'), 8192, free_space
+    ).tap do |resource|
+      allow(resource).to receive(:accessible?).and_return(true)
+    end
+  end
```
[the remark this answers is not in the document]
ktchen14: After talking to Mikhail about this issue I'll retract my earlier comment. The clarity of this spec will improve if we use an `instance_double` for `VSphereCloud::Resources::Datastore` here as we can drop several of the magic values we're using and limit the scope of the unit test to the actual `VMPlacementSelectionPipeline` unit.

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_placement_selection_pipeline_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+  let(:ds_shared) { fake_datastore('fake-ds-shared') }
+  let(:cluster_1) { fake_cluster('fake-cluster-1', ds_cl1_1, ds_cl1_2, ds_shared) }
+  let(:cluster_2) { fake_cluster('fake-cluster-2', ds_cl2_1, ds_cl2_2, ds_shared) }
+  let(:placement_1) { fake_placement(cluster_1) }
+  let(:placement_2) { fake_placement(cluster_2) }
+  let(:disk_config) {[]}
+
+  context 'when there are no disks to be configured' do
+    it 'only generates placements with enough memory' do
+      allow_any_instance_of(VSphereCloud::Resources::Datastore).to receive(:maintenance_mode?).and_return(false)
```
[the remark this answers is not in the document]
EleanorRigby: I would rather keep it as a single context then. maintenance-mode? of datastores is just a filter and making a different context for it would be confusing.
Right now I have arranged it in a way where we either configure 0 disks and concentrate on all other filters or have 3 disks.

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_placement_selection_pipeline_spec.rb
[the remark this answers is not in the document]
medvedzver: I see multiple reasons for changing. To be consistent with unit test approach. Where we don't need to rely on what's going on in datastore, we just need to mock it.
Another thing, using instance double shows explicitly what fields you are using. Right now I see 2 params that don't make sense to me: true and 8192. 
The same goes to `fake_placement`.

src/vsphere_cpi/spec/unit/cloud/vsphere/selection_pipeline_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+
+    context 'with no block' do
+      it 'raises an ArgumentError when no arguments are given' do
+        expect { subject.with_scorer }.to raise_error(ArgumentError)
+      end
+
+      it 'adds each argument as a scorer' do
+        subject.with_scorer(:fake_1, :fake_2)
+        list = subject.send(:scorer_list)
+        expect(list).to include(:fake_1, :fake_2)
```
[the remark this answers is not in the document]
EleanorRigby: scorer_list is a private method and send is the only way to call it. but i can make it smaller by moving out the list var

src/vsphere_cpi/spec/unit/cloud/vsphere/selection_pipeline_spec.rb
```diff
@@ -0,0 +1,170 @@
...
+
+      filter_3 = double
+      expect(filter_3).to receive(:call).with(:fake, criteria).and_return(false)
+      subject.with_filter(filter_3)
+
+      filter_4 = double
+      expect(filter_4).not_to receive(:call)
+      subject.with_filter(filter_4)
+
+      subject.accept?(:fake)
```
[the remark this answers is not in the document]
EleanorRigby: This is just to trigger calling of each filter. The test wants to make sure that each filter in the list is called until a false (rejection by a filter) is encountered.

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
EleanorRigby: I do not understand the intent completely. Can this be taken up in a separate story to work on scorers?
I want to design them in a way to add some logging too.

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
EleanorRigby: The design in such that any logging here will generate nC2(combination) messages for n placements. The info printed with each selected placement should be sufficient to infer for now. We can discuss on changing scoring design later.

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_placement_selection_pipeline_spec.rb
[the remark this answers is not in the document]
EleanorRigby: I have another thing to add here. How can I control behavior of all datastores that I am using with just one line. For Instance, if I use Resources instead of instance_double I can say something like 
```
allow_any_instance_of(VSphereCloud::Resources::Datastore).to receive(:maintenance_mode?).and_return(true)
```
I cannot do this on instance_double or is there a way?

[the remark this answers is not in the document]
EleanorRigby: Since I am dealing with multiple datastores here, I want to control their behavior in bulk. Does this make sense?

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
ktchen14: The problem is that unfortunately some of the code in `VmPlacementSelectionPipeline` is just wrong:

```ruby
    with_scorer do |p1, p2|
      p1.migration_size <=> p2.migration_size
    end

    with_scorer do |p1, p2|
      p1.free_memory <=> p2.free_memory
    end
```

Notice that `migration_size` and `free_memory` are both used to determine the scoring for `VmPlacement`s. However these two metrics have an inverse relationship: **lower** `migration_size` is good while **higher** `free_memory` is good.

This means that either:

1. The `migration_size` scorer is wrong here and needs to be:

   ```ruby
   p2.migration_size <=> p1.migration_size
   ```

2. Or the `free_memory` scorer is wrong here and needs to be:

   ```ruby
   p2.free_memory <=> p1.free_memory
   ```

Unfortunately I can't tell which one is wrong which is why I think that we need to rethink `compare_placements`.

That being said we can merge this and fix later if you want to avoid blocking the other tracks of work.

EleanorRigby:
I have a quick and clean fix to it. It is not the best fix but will make it
behave like expected. Will do it today

src/vsphere_cpi/lib/cloud/vsphere/selection_pipeline.rb
EleanorRigby: So,  

`p1.migration_size <=> p2.migration_size` sorts in ascending order but 

`-(p1.migration_size <=> p2.migration_size)` should sort in descending order.

Having a negative sign should make it easier to write each scorer. 

But I agree that when to use negative and when not to use negative is not very intuitive while writing and can lead to bugs. 

For now it seems to be a quick and good fix. Thoughts?

nehagjain15: this bug should have resulted in a test failure right?

nehagjain15: just saw the test changes

EleanorRigby: yeah, it should have failed. Test was wrong. It expected everything to be in descending order.
Now it is corrected.

src/vsphere_cpi/lib/cloud/vsphere/disk_placement_selection_pipeline.rb
```diff
@@ -34,27 +39,38 @@ def ==(other)
...
     # Select storage placements which match target datastore pattern
     with_filter do |storage_placement, criteria_object|
+      logger.debug("Filter #{storage_placement.name} for target DS pattern #{criteria_object.target_datastore_pattern}")
       storage_placement.name =~ Regexp.new(criteria_object.target_datastore_pattern)
     end
 
     # Select storage placements that are accessible from any host
-    with_filter { |storage_placement| storage_placement.accessible? }
+    with_filter do |storage_placement|
+      logger.debug("Filter #{storage_placement.name} for free accessibility")
```
nehagjain15: remove 'free'

EleanorRigby: good catch

PR #109, opened 2018-05-31
[WIP] GPU Support in VMs

EleanorRigby:
This PR adds support in vSphere CPI to attach GPU on a Esxi host as a passthrough devices onto a VM.
  1. It currently assumes that there are no anti-affinity DRS rules set up on any cluster thus selection of a host (with GPU) for placing a vm will not fail while cloning for any affinity related reason.
  2. It only supports GPU in passthrough(direct) mode.
  3. It handles scenarios where some hosts in a cluster have enough available GPU but are not accessible by any datastore (otherwise accessible from the other hosts in the cluster).
  4. It supports specifying number of GPUs required by a BOSH VM


EleanorRigby:
Logic for filtering placements is in place. Logic for selecting a placement and reconfiguration of vm post creation is WIP

src/vsphere_cpi/lib/cloud/vsphere/cluster_picker.rb
```diff
@@ -67,14 +72,55 @@ def best_cluster_placement(req_memory:, disk_configurations:)
...
+          # Add this list to host_gpu_map with host name as the key unless this list is empty
+          placements[cluster.name][:host_gpu_map][host.name] = available_gpus unless available_gpus.empty?
+        end
+        # Remove the placement if it has no such hosts
+        placements.delete(cluster.name) if placements[cluster.name][:host].empty?
+      end
+      placements
+    end
+
+    def available_gpus(host_mob, num_gpu_required)
```
cppforlife: isnt this method racy since multiple vms in an instance group could be created in parallel?

EleanorRigby: It is  racy but so is the selection of datastore and cluster. CPI has no control over it since for each create_vm call bosh starts a unique CPI process. 
I can randomize the selection of host(with GPUs) from a cluster like we do for datastores. Beyond that, The only way I see a solution is to add random sleeps in create_vm to avoid races.
@yeshwantbabar : we need to discuss this.

EleanorRigby: This feature is experimental and has few more assumptions in the current form; like, breaking drs affinity rules.
For now, I think best way will be to create a separate vm_type with GPU specs and create limited instances with GPUs in a separate cluster. Since each create_vm will see same number of gpus available on each host of the cluster, it will be best to have enough gpus on each host.

Issue #108, opened 2018-05-31
Datastore disk placement issue in v47 


brunograz:
Hi,

We deployed a bosh director specifying the parameters listed at https://bosh.io/docs/init-vsphere/. What we noticed though was a strange behavior when deploying VMs. 
Every single ephemeral disk would use the same datastore until it fills up and return us the error `Error: Unknown CPI error 'Unknown' with message 'Module 'MonitorLoop' power on failed. ' in 'create_vm' CPI method`.

Following the documentation, we tried multiple configurations. From changing the regex in the init command to specifying the datastores/ds_cluster for persistent and ephemeral disks in the cloud config as per https://bosh.io/docs/vsphere-cpi/#resource-pools. None of those options worked and we were seeing the same behavior all the time.

Our final 'regex' looked like: `^(DS0|DS1|DS2|DS3|DS4|DS5|DS6)$` and even then only a single datastore was picked every time. 

The storage config comprises of a **single** datastore cluster with 6 datastores in it, SDRS is enabled. 
Every physical host has the permission to access the 6 datastores.

Interestingly, changing the cpi version from v47 to v45.1.0 and using the same regex as shown above resulted in the expected disk placement as for https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/blob/master/src/vsphere_cpi/lib/cloud/vsphere/datastore_picker.rb#L14. 

v47 has added support for SDRS and I wonder if that has impact on what was described above - is there an additional configuration required?. 

We didn't check if the same applies for v48 or v49.

nehagjain15:
@brunograz In in your config did you only specify the Datastore cluster which had those 6 datastores?
Could you share what was in your init command and in cloud config please.  
If you have the logs that would be of great help, if you could share those, so that we can see what cpi did. You can get cpi specific log from the deploy task using --cpi flag. If not do you mind retrying with v47 and then sharing the logs as that would help narrow down the bug quickly. v47 has SDRS support but it should not affect your config if you don't specify clusters in it. If you want to use SDRS support you can specify your datastores in cloud config as datastores: [{clusters: [datastrore_cluster_name1: {}  , datastore_cluster_name2: {}]}]

nehagjain15:
@brunograz we did find what the issue is and will be fixing it. Tracker [link](https://www.pivotaltracker.com/story/show/158039615)
[...]

PR #111, opened 2018-06-02
Fix datastore picker logic

EleanorRigby:
- Filter inaccessible datastore (not accessible from any host) before
calling datastore picker.
- Change cluster_picker, create_disk and reattach_disk to filter as per
above criteria.
- Datastore picker now only selects one possible datastore for each disk.
- Fix Unit tests
- Fix Integration tests

#108 

[Delivers #158039615](https://www.pivotaltracker.com/story/show/158039615)

Signed-off-by: Neha Jain <jneha@vmware.com>

### DIFF
--- a/src/vsphere_cpi/spec/unit/cloud/vsphere/cluster_picker_spec.rb
+++ b/src/vsphere_cpi/spec/unit/cloud/vsphere/cluster_picker_spec.rb
@@ -67,7 +67,7 @@ module VSphereCloud
 
         context 'when all hosts are in maintenance mode' do
           let(:host_runtime_info) { instance_double(VimSdk::Vim::Host::RuntimeInfo, in_maintenance_mode: true) }
-          it 'returns the no active host exception' do
+          it 'raises the error' do
             disks = [
                 instance_double(VSphereCloud::DiskConfig,
                                 size: 256,
@@ -87,7 +87,7 @@ module VSphereCloud
 
             expect do
               picker.best_cluster_placement(req_memory: 1024, disk_configurations: disks)
-            end.to raise_error(/No valid placement found due to no active host/)
+            end.to raise_error(/No valid placement found/)
           end
         end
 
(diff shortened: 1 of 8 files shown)


src/vsphere_cpi/spec/unit/cloud/vsphere/cluster_picker_spec.rb
```diff
@@ -85,9 +97,9 @@ def fake_datastore(name, free_space, mob = nil)
             picker = ClusterPicker.new(0, 0)
             picker.update(available_clusters)
 
-            expect do
+            expect {
               picker.best_cluster_placement(req_memory: 1024, disk_configurations: disks)
-            end.to raise_error(/No valid placement found/)
+            }.to raise_error(/No valid placement found/)
```
EleanorRigby: What difference does it cause to use { vs do

src/vsphere_cpi/spec/unit/cloud/vsphere/datastore_picker_spec.rb
```diff
@@ -378,6 +378,10 @@ module VSphereCloud
           sorted_datastores = available_datastores.sort do |x, y|
             y[1].free_space <=> x[1].free_space
           end
+          puts 'Summary of simulated placements'
```
EleanorRigby: I removed this earlier. Why do we need to print this? Is it to verify the randomization taking place?

nehagjain15: yes, Like earlier it was 400 for each datastore since each disk was going into each datastore. We could think of may be adding a check instead of printing if those no's are different when we refactor

src/vsphere_cpi/spec/unit/cloud/vsphere/cluster_picker_spec.rb
nehagjain15: nothing, I just did it for consistency in this spec file

PR #112, opened 2018-06-05
Make setup_env_proxy immune to close ssh socket failure

EleanorRigby:
- On a concourse container it is increasingly noticed that ssh socket
connection fails to close cleanly and results in script exiting with non
zero status due to set -e flag. This commit disables the set -e and
forces the ssh socket closure command to return 0 status.


EleanorRigby:
okay I will try this on refactor-disk dummy pipeline

PR #116, opened 2018-06-14
[WIP] Adds diagnostic script for cluster & ds picker

EleanorRigby:
- The diagnostic script prints
  - CPI datacenter info
  - Takes in additional cluster as argumnets (aside from what cloud
  config has)
  - Runs cluster picker algorithm 1000 times, it does not mean VMs are
  actually getting created and space on ds is not actually filling up.
  - Each run of cluster picker sees clusters and datastores exactly
  same as any other run of cluster picker.

### DIFF
--- a/src/vsphere_cpi/bin/run_diagnose
+++ b/src/vsphere_cpi/bin/run_diagnose
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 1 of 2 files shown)


EleanorRigby:
This is for discussing and building a diagnostic script for picker logic.

TODO : Allow user to specify datastore patterns as options for command line

EleanorRigby:
Although with refactoring of picker coming in sometime soon this script will change considerably.

PR #125, opened 2018-07-11
account for swapfiles when choosing a datastore

### LINKED ISSUE #98: Free disk space calculation in datastore_picker.rb doesn't take vmware swapfiles into account
There is a hardcoded value for the headroom in the datastore_picker.rb that's used when selecting datastores to place a vm/disk. The CPI's choice is based on free space + headroom value but unfortunately it doesn't take into account the fact that vmware creates a temporary swapfile (same size as the vm-RAM is). This might lead to the situation that bosh tries to create this swapfile on a datastore which has not enough free space to do so. You'll see an error message like this:

```
Error: Unknown CPI error 'Unknown' with message 'Failed to extend swap file from 0 KB to XXXX KB
```

We've solved this issue by temporarily changing the headroom value to at least the size of the biggest instance (RAM wise) in the vsphere-cpi code directly on the bosh-director.

Could you please fix this bug by changing the free space calculation for datastores to account for the additional needed vm-RAM size? It would be nice if this is somehow similar handled as it is done for the free disk space calculation by adding the headroom value.

kreamyx:
fixes #98

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cluster_picker.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cluster_picker.rb
@@ -30,7 +30,7 @@ module VSphereCloud
         datastore_picker.update(accessible_datastores)
 
         begin
-          placement = datastore_picker.best_disk_placement(disk_configurations)
+          placement = datastore_picker.best_disk_placement(disk_configurations, max_swapfile_size=req_memory)
           placement[:memory] = cluster.free_memory
           [cluster.name, placement]
         rescue Bosh::Clouds::CloudError
--- a/src/vsphere_cpi/lib/cloud/vsphere/cluster_picker.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cluster_picker.rb
@@ -30,7 +30,7 @@ module VSphereCloud
         datastore_picker.update(accessible_datastores)
 
         begin
-          placement = datastore_picker.best_disk_placement(disk_configurations, max_swapfile_size=req_memory)
+          placement = datastore_picker.best_disk_placement(disk_configurations, max_swapfile_size: req_memory)
           placement[:memory] = cluster.free_memory
           [cluster.name, placement]
         rescue Bosh::Clouds::CloudError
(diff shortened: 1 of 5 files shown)


src/vsphere_cpi/lib/cloud/vsphere/datastore_picker.rb
```diff
@@ -18,7 +18,7 @@ def update(available_datastores={})
     # This means the largest datastores have the highest chance of being first, but adds a bit of randomization.
     # After the sort, disks are placed into the first datastore that has enough space for that disk.
     #
-    def best_disk_placement(disks)
+    def best_disk_placement(disks, max_swapfile_size=0)
```
ktchen14: https://github.com/rubocop-hq/ruby-style-guide#spaces-around-equals

ktchen14: Also `max_swapfile_size` is probably better as a keyword argument

src/vsphere_cpi/lib/cloud/vsphere/cluster_picker.rb
```diff
@@ -30,7 +30,7 @@ def best_cluster_placement(req_memory:, disk_configurations:)
         datastore_picker.update(accessible_datastores)
 
         begin
-          placement = datastore_picker.best_disk_placement(disk_configurations)
+          placement = datastore_picker.best_disk_placement(disk_configurations, max_swapfile_size=req_memory)
```
ktchen14: This doesn't exactly do what I think you're trying to do; specifically it:

1. Declares a new local variable `max_swapfile_size`
2. Assigns `req_memory` to `max_swapfile_size`
3. Passes the result of the assignment as the second argument

Mostly it's the same result but unnecessarily creates a variable. Ruby has no direct equivalent of Python's ability to assign any function argument by name with `=`.

kreamyx: Got you! I will just use `req_memory` as a second parameter then

PR #129, opened 2018-07-30
Account for swapfiles when choosing a datastore

### LINKED ISSUE #98: Free disk space calculation in datastore_picker.rb doesn't take vmware swapfiles into account

EleanorRigby:
fixes #98

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/disk_placement_selection_pipeline.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/disk_placement_selection_pipeline.rb
@@ -2,20 +2,36 @@ require 'forwardable'
 
 module VSphereCloud
   class DiskPlacementSelectionPipeline < SelectionPipeline
-    DISK_HEADROOM = 1024
-
     class StoragePlacementCriteria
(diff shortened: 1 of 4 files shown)


EleanorRigby:
@kreamyx

kreamyx:
@EleanorRigby lgtm

PR #131, opened 2018-08-01
Enable CPI extensions

EleanorRigby:
[#157263518](https://www.pivotaltracker.com/story/show/157263518)

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -2,6 +2,7 @@ require 'cloud/vsphere/logger'
 require 'json'
 require 'membrane'
 require 'cloud'
+require 'cloud/vsphere/cpi_extension'
 
 module VSphereCloud
   class Cloud < Bosh::Cloud
@@ -138,6 +139,7 @@ module VSphereCloud
       with_thread_name("create_stemcell(#{image}, _)") do
         # Add cpi telemetry advanced config to vc in a separate thread
         telemetry_thread = Thread.new { enable_telemetry }
+        VCPIExtension.create_cpi_extension(client)
         result = nil
         Dir.mktmpdir do |temp_dir|
           logger.info("Extracting stemcell to: #{temp_dir}")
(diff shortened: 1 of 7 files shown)


EleanorRigby:
@nehagjain15 : do you think it will be good idea to move all extension manager stuff in a separate module? I am not liking the way I am putting all things in cloud.rb. Thoughts?

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -41,6 +44,44 @@ def enable_telemetry
       other_client.logout rescue nil
     end
 
+    def enable_extension
+      http_client = VSphereCloud::CpiHttpClient.new
```
nehagjain15: do you need a separate client for this?

nehagjain15:
That sounds great to move it out from cloud.rb

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -41,6 +44,44 @@ def enable_telemetry
       other_client.logout rescue nil
     end
 
+    def enable_extension
```
nehagjain15: Should this method be named create_cpi_extension instead as you are either returning and existing/creating a new extension right?

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
EleanorRigby: Just for the fact that it is a separate operation like adding telemetry configuration and keeping it asynchronous helps to speed up things. It takes sometime to register and find an extension.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
EleanorRigby: yep, I can rename that.

src/vsphere_cpi/lib/cloud/vsphere/cpi_extension.rb
```diff
@@ -0,0 +1,69 @@
...
+
+    def register_extension(extension)
+      ext_mgr = vc_client.service_content.extension_manager
+      ext_mgr.register_extension(extension)
+    end
+
+    def extension_is_registered?(key: DEFAULT_VSPHERE_CPI_EXTENSION_KEY)
+      ext_mgr = vc_client.service_content.extension_manager
+      cpi_extension = ext_mgr.find_extension(key)
+      logger.debug("Extension already exists")
```
nehagjain15: shouldn't this log only if !cpi_extencion.nil? else it would be misleading

PR #132, opened 2018-08-01
Nj update pyvmomi 153524210

nehagjain15:


src/vsphere_cpi/lib/ruby_vim_sdk/vmodl/managed_object.rb
```diff
@@ -40,7 +40,7 @@ def finalize
             self.instance_eval do
               define_method(method.name) do |*args|
                 unless args.length.between?(method.arguments.length - method.num_optional_args, method.arguments.length)
-                  raise ArgumentError, "wrong number of arguments (#{args.length} for #{method.arguments.length} with optional args #{method.number_of_optional_args_in_end})"
+                  raise ArgumentError, "wrong number of arguments (#{args.length} for #{method.arguments.length} with optional args #{method.num_optional_args})"
                 end
```
EleanorRigby: Didn't we correct this as part of some previous commit?

src/vsphere_cpi/lib/ruby_vim_sdk/soap/stub_adapter.rb
```diff
@@ -32,9 +33,9 @@ def invoke_method(managed_object, method_info, arguments, outer_stub = nil)
           elsif status == 200
             result = object
           elsif object.kind_of?(Vmodl::MethodFault)
-            raise SoapError.new(object.msg, object)
+            raise VimSdk::SoapError.new(object.msg, object)
```
EleanorRigby: Is it necessary to use `VimSdk::SoapError` instead of just `SoapError` as we are already inside `VimSdk` module.

scripts/pyvmomi_to_ruby/QueryTypes.py
```diff
@@ -1,63 +0,0 @@
-# ******* WARNING - AUTO GENERATED CODE - DO NOT EDIT *******
```
EleanorRigby: Is there any reason to delete the file now? or rather why this file was not fetched from url earlier?

nehagjain15: It was fetched from url but we don't commit the .py files and I wasn't aware about this when I upgraded pyvmomi to earlier version

src/vsphere_cpi/lib/ruby_vim_sdk/soap/stub_adapter.rb
nehagjain15: Yes its needed.

PR #134, opened 2018-08-09
Chore/use httprb client

medvedzver:


src/vsphere_cpi/lib/nsxt/nsxt_client/api_client.rb
```diff
@@ -48,125 +38,94 @@ def self.default
...
+        fail ApiCallError.new(:code => response.code,
+                              :response_headers => response.headers,
+                              :response_body => response.body,
+                              :message => error_message),
+             response.status
       end
 
       if opts[:return_type]
-        data = deserialize(response, opts[:return_type])
+        data = deserialize(response.to_s, opts[:return_type])
```
nehagjain15: why are we doing to_s on response? should this be response.body

medvedzver: `response.to_s` gives you a response body.

EleanorRigby:
NSX-T, moving to Policy API  will make the changes to the manager world unnecessary.

Similar choice can be made to use httprb in policy swagger client as and when required. Closing it.

PR #139, opened 2018-08-17
[WIP]Enable Storage Policy based Management for disk encryption

nehagjain15:
DO NOT MERGE


src/vsphere_cpi/lib/ruby_vim_sdk/soap/stub_adapter.rb
```diff
@@ -23,16 +22,20 @@ def invoke_method(managed_object, method_info, arguments, outer_stub = nil)
...
+        response = @http_client.post(@api_uri, request, headers)
 
         status = response.code
         if status == 200 || status == 500
           object = SoapResponseDeserializer.new(outer_stub).deserialize(response.content, method_info.result_type)
           if outer_stub != self
+            @vc_cookie = response.cookies.first.value if response.cookies && @vc_cookie.nil?
             result = [status, object]
           elsif status == 200
+            @vc_cookie = response.cookies.first.value if response.cookies && @vc_cookie.nil?
```
EleanorRigby: Why `@vc_cookie.nil?` ?

nehagjain15: do not want to overwrite the vc_cookie value.. I don't remember the exact details but I think the cookie value changes later on with other requests and is set the first time when you login to VC.

PR #146, opened 2018-09-07
Initialize migration size to zero

EleanorRigby:
Add Unit tests
Add Integration Test to excercise existing persistent disk scenario

### DIFF
--- a/src/vsphere_cpi/spec/integration/vm_existing_disk.rb
+++ b/src/vsphere_cpi/spec/integration/vm_existing_disk.rb
@@ -1,0 +1,84 @@
+require 'integration/spec_helper'
+
+describe 'Creating VM with existing persistent disks' do
+  before (:all) do
+    @datacenter_name = fetch_and_verify_datacenter('BOSH_VSPHERE_CPI_DATACENTER')
+    @cluster_name = fetch_and_verify_cluster('BOSH_VSPHERE_CPI_CLUSTER')
+    @datastore_shared_pattern = fetch_and_verify_datastore('BOSH_VSPHERE_CPI_SHARED_DATASTORE', @cluster_name)
+    @cluster_one_datastore = fetch_and_verify_datastore('BOSH_VSPHERE_CPI_DATASTORE_PATTERN', @cluster_name)
+  end
+
+  context 'when there are existing persistent disks that needs to be migrated to new datastore' do
+    let(:vm_type) do
+      {
+        'ram' => 512,
+        'disk' => 2048,
+        'cpu' => 1
+      }
+    end
+    let(:cpi) do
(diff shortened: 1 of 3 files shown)


src/vsphere_cpi/spec/integration/vm_existing_disk.rb
```diff
@@ -0,0 +1,84 @@
...
+
+describe 'Creating VM with existing persistent disks' do
+  before (:all) do
+    @datacenter_name = fetch_and_verify_datacenter('BOSH_VSPHERE_CPI_DATACENTER')
+    @cluster_name = fetch_and_verify_cluster('BOSH_VSPHERE_CPI_CLUSTER')
+    @datastore_shared_pattern = fetch_and_verify_datastore('BOSH_VSPHERE_CPI_SHARED_DATASTORE', @cluster_name)
+    @cluster_one_datastore = fetch_and_verify_datastore('BOSH_VSPHERE_CPI_DATASTORE_PATTERN', @cluster_name)
+  end
+
+  context 'when there are existing persistent disks ' do
```
nehagjain15: Creating VM with existing persistent disks 'when disks needs to be migrated to new datastore ' cpi should be able to create a VM. I think context needs to be updated to make clear what you are actually testing

src/vsphere_cpi/spec/integration/vm_existing_disk.rb
```diff
@@ -0,0 +1,84 @@
...
+        vm_id = cpi_different_persistent_pattern.create_vm(
+          'agent-007',
+          @stemcell_id,
+          vm_type,
+          get_network_spec,
+          [@disk_id_one, @disk_id_two],
+          {}
+        )
+        expect(vm_id).to_not be_nil
+        expect(cpi_different_persistent_pattern.has_vm?(vm_id)).to be(true)
```
nehagjain15: you could add another expectation here that the disk is in new datastore

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_placement_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'spec_helper'
+
+describe VSphereCloud::VmPlacement do
```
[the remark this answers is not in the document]
nehagjain15: Agree

src/vsphere_cpi/spec/integration/vm_existing_disk.rb
EleanorRigby: Disk not in new datastore. it will be when `attach_disk` is called. That never happens now as we are in integration test, but BOSH will likely do it

PR #147, opened 2018-09-07
Add extension to replicated stemcell

EleanorRigby:
- Add Integration test
- Fix Unit Tests

### DIFF
--- a/src/vsphere_cpi/spec/integration/cpi_extension_spec.rb
+++ b/src/vsphere_cpi/spec/integration/cpi_extension_spec.rb
@@ -3,6 +3,11 @@ require 'integration/spec_helper'
 require 'cloud/vsphere/cpi_extension'
 
 describe '#add cpi extension' do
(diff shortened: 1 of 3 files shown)


src/vsphere_cpi/spec/integration/cpi_extension_spec.rb
```diff
@@ -41,6 +44,40 @@
...
+        datacenters: [{
+          clusters: [@cluster_name],
+          datastore_pattern: @datastore_pattern,
+          persistent_datastore_pattern: @datastore_pattern
+        }],
+      )
+      VSphereCloud::Cloud.new(options)
+    end
+
+    it 'adds replicated stemcell to extension' do
```
nehagjain15: 'add replicated stemcell and vm to extension' 
since you are testing that vm also is added to extension in this test, you can essentially delete this test which is now redundant  'adds vm to the default cpi extension''

PR #160, opened 2018-10-09
Remove swagger spec patching

EleanorRigby:
# Description
Swagger Spec used before was patched to add a missing operation. Upon
investigation, operation was not missing and wrong operation got
introduced due to applying of patch.

## Type of change

## How Has This Been Tested?
https://ci.vcna.io/teams/vcpi/pipelines/vcpi-ta-nsxtRefactor

# Checklist:

- [NA] My code follows the standard ruby style guide
- [NA] I have commented my code, particularly in hard-to-understand areas
- [NA] I have added tests that prove my fix is effective or that my feature works


src/vsphere_cpi/lib/nsxt/nsxt_client/api/logical_routing_and_services_api.rb
```diff
@@ -1179,57 +1179,6 @@ def delete_logical_router_port_with_http_info(logical_router_port_id, opts = {})
       return data, status_code, headers
     end
 
-    # Delete a Logical Router Port
-    # Deletes the specified logical router port. You must delete logical router ports before you can delete the associated logical router. To Delete Tier0 router link port you must have to delete attached tier1 router link port, otherwise pass 'force=true' as query param to force delete the Tier0 router link port.  
-    # @param [Hash] opts the optional parameters
-    # @option opts [BOOLEAN] :force Force delete the resource even if it is being used somewhere  (default to false)
-    # @return [nil]
-    def delete_logical_router_ports(opts = {})
```
nehagjain15: Shouldn't this be there and the patched delete_logical_router_port() method be delete? also you will have to update the code to call this method instead of the patched one

EleanorRigby: So this is not supposed to be there. 

` delete_logical_router_port() ` is there and is being called in code.

PR #170, opened 2018-11-06
Enable host group specification in a Cluster

EleanorRigby:
# Description

This PR enables an Admin to specify a host group in the vSphere clusters. 

This directly enables the Admin to define multiple AZs within the same cluster by using host groups. 

VM Placement & Creation

## Type of change

## How Has This Been Tested?

Unit and Integration tests
# Checklist:

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cluster_config.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cluster_config.rb
@@ -8,8 +8,8 @@ module VSphereCloud
       @config = config_hash
     end
 
-    def resource_pool
-      @config['resource_pool']
-    end
+    def host_group; @config['host_group']; end
+
+    def resource_pool; @config['resource_pool']; end
   end
 end
--- a/src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
@@ -5,7 +5,7 @@ module VSphereCloud
     include Logger
 
     CUSTOM_ATTRIBUTE_NAME = 'drs_rule'
-    DRS_LOCK_SUFFIX_VMGROUP = '_vm_group'
+    DRS_LOCK_HOST_VM_GROUP = 'host_vm_group'
     DEFAULT_RULE_NAME = 'vcpi-drs-rule'
 
     def initialize(rule_name, client, datacenter_cluster)
@@ -28,6 +28,22 @@ module VSphereCloud
           update_rule(rule.key)
         else
           add_rule
+        end
+      end
+    end
+
+    # Adds VM to given VM Group & creates VM/HOST Affinity rule
+    # @param [Vim::VirtualMachine] vm
+    # @param [String] vm_group_name
+    # @param [String] host_group_name
+    def add_vm_host_affinity_rule(vm_group_name, host_group_name)
+      DrsLock.new(@vm_attribute_manager, DRS_LOCK_HOST_VM_GROUP).with_drs_lock do
+        rule = find_rule
+        # Do not create the rule if it already exists
+        unless rule
+          # No error is raised if given host group does not exist,
+          # it still creates the rule
+          create_vm_host_affinity_rule(vm_group_name, host_group_name)
         end
       end
     end
@@ -57,6 +73,26 @@ module VSphereCloud
     def update_rule(rule_key)
       logger.debug("Updating DRS rule: #{@rule_name}")
       add_anti_affinity_rule(VimSdk::Vim::Option::ArrayUpdateSpec::Operation::EDIT, rule_key)
+    end
+
+    def create_vm_host_affinity_rule(vm_group_name, host_group_name)
+      vm_host_rule_info = VimSdk::Vim::Cluster::VmHostRuleInfo.new
+      vm_host_rule_info.enabled = true
+      # This is a HARD AFFINITY rule.
+      vm_host_rule_info.mandatory = true
+      vm_host_rule_info.name = @rule_name
+      vm_host_rule_info.vm_group_name = vm_group_name
+      vm_host_rule_info.affine_host_group_name = host_group_name
+
+      cluster_rule_spec = VimSdk::Vim::Cluster::RuleSpec.new
+      cluster_rule_spec.info = vm_host_rule_info
+      cluster_rule_spec.operation = VimSdk::Vim::Option::ArrayUpdateSpec::Operation::ADD
+
+      config_spec = VimSdk::Vim::Cluster::ConfigSpecEx.new
+      config_spec.rules_spec = [cluster_rule_spec]
+
+      logger.debug("Creating VM/Host Affinity Rule: #{@rule_name}")
+      reconfigure_cluster(config_spec)
     end
 
     def add_anti_affinity_rule(operation, rule_key = nil)
@@ -96,3 +132,4 @@ module VSphereCloud
     end
   end
 end
+
--- a/src/vsphere_cpi/lib/cloud/vsphere/drs_rules/vm_group.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/drs_rules/vm_group.rb
@@ -17,7 +17,7 @@ module VSphereCloud
     # @param [Vim::VirtualMachine] vm
     # @params [String] vm_group_name
     def add_vm_to_vm_group(vm, vm_group_name)
-      DrsLock.new(@vm_attribute_manager, DRS_LOCK_VMGROUP ).with_drs_lock do
+      DrsLock.new(@vm_attribute_manager, DRS_LOCK_VMGROUP).with_drs_lock do
         vm_group = find_vm_group(vm_group_name)
         logger.debug("VmGroup: #{vm_group_name} already exists") unless vm_group.nil?
 
@@ -43,7 +43,7 @@ module VSphereCloud
     def delete_vm_groups(vm_group_names)
       return if vm_group_names.empty?
       vm_attribute_manager = VMAttributeManager.new(@client.service_content.custom_fields_manager)
-      DrsLock.new(vm_attribute_manager, DRS_LOCK_VMGROUP ).with_drs_lock do
+      DrsLock.new(vm_attribute_manager, DRS_LOCK_VMGROUP).with_drs_lock do
         empty_vm_groups = @cluster.configuration_ex.group.select do |group|
           group.is_a?(VimSdk::Vim::Cluster::VmGroup) && vm_group_names.include?(group.name) && group.vm.empty?
         end
@@ -59,8 +59,22 @@ module VSphereCloud
           group_spec.remove_key = vm_group.name
           group_spec
         end
+
+        # If these vm_groups are getting deleted. Also delete all the
+        # vm-host group rules associated with it.
+        #
+        # @TODO : CAVEAT: if anyone creates a rule outside of CPI using these
+        # vm_groups, it will be deleted too. To narrow
+        # the scope of this caveat, CPI will only delete the
+        # vm-host-group-affinity rule type. This is because, currently it only
+        # creates vm-host-affinity rule using vm groups.
+        rules_spec = get_delete_rule_spec(empty_vm_groups) unless empty_vm_groups.nil?
         config_spec = VimSdk::Vim::Cluster::ConfigSpecEx.new
         config_spec.group_spec = group_specs
+        unless rules_spec&.empty?
+          logger.debug("Deleting #{rules_spec.count} vm host affinity rules")
+          config_spec.rules_spec = rules_spec
+        end
         @client.wait_for_task do
           @cluster.reconfigure_ex(config_spec, true)
         end
@@ -68,6 +82,29 @@ module VSphereCloud
     end
 
     private
+    def get_delete_rule_spec(empty_vm_groups)
+      # returning empty array for consistent return type.
+      if @cluster.configuration_ex.rule.nil?
+        return []
+      end
+      rules_spec= empty_vm_groups.each_with_object([]) do |vm_group, rules_spec|
+        # Filter rules for deletion
+        rule_keys = @cluster.configuration_ex.rule.select do |rule_info|
+          rule_info.is_a?(VimSdk::Vim::Cluster::VmHostRuleInfo) && \
+              rule_info.vm_group_name == vm_group.name
+        end.map(&:key)
+
+        # Generate Rule Spec
+        rule_keys.each do |rkey|
+          rule_spec = VimSdk::Vim::Cluster::RuleSpec.new
+          rule_spec.operation = VimSdk::Vim::Option::ArrayUpdateSpec::Operation::REMOVE
+          rule_spec.remove_key = rkey
+          rules_spec << rule_spec
+        end
+      end
+      rules_spec
+    end
+
 
     def find_vm_group(vm_group_name)
       @cluster.configuration_ex.group.find do |group|
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
@@ -9,6 +9,8 @@ module VSphereCloud
       include Logger
       stringify_with :name
 
+      CLUSTER_VM_GROUP_SUFFIX = '_vm_group'.freeze
+      CLUSTER_VM_HOST_RULE_SUFFIX = '_rule'.freeze
       PROPERTIES = %w(name datastore resourcePool host)
       HOST_PROPERTIES = %w(name hardware.memorySize runtime.connectionState runtime.inMaintenanceMode runtime.powerState)
       HOST_COUNTERS = %w(mem.usage.average)
@@ -45,10 +47,65 @@ module VSphereCloud
         # Have to use separate mechanisms for fetching utilization depending on
         # whether we're using resource pools or raw clusters.
         if @config.resource_pool.nil?
-          @synced_free_memory = fetch_cluster_utilization
+          # Get cluster utilization only if host group is nil
+          @synced_free_memory = if host_group.nil?
+            fetch_cluster_utilization
+          else
+            fetch_host_group_utilization
+          end
         else
           @synced_free_memory = fetch_resource_pool_utilization
         end
+      end
+
+      # Returns a default name made from host group and vm group.
+      #
+      # @return [String] Name of the host VM affinity rule name.
+      def vm_host_affinity_rule_name
+        # This first return might never get invoked as we always check host group
+        # to be present before calling this function.
+        # Just in case somebody uses this function in future without checking
+        # host group, this function will be safe.
+        return nil if host_group.nil?
+        vm_group + CLUSTER_VM_HOST_RULE_SUFFIX
+      end
+
+      # Returns a host group name if specified in cluster config. Otherwise,
+      # it returns a nil.
+      #
+      # @return [String] Name of the host Group specified under a cluster config.
+      def host_group
+        config.host_group
+      end
+
+      # it returns a default name created by adding suffix to host group name
+      #
+      # @return [String] Name of the VM Group
+      def vm_group
+        return nil if host_group.nil?
+        return "vm_group-#{(0...6).map { (97 + rand(26)).chr }.join}" if host_group.length > 100
+        host_group + CLUSTER_VM_GROUP_SUFFIX
+      end
+
+      def host_group_mob
+        return nil if host_group.nil?
+        result_host_group = mob.configuration_ex.group.find do |group|
+          group.name == host_group && group.is_a?(VimSdk::Vim::Cluster::HostGroup)
+        end
+        raise "Failed to find the HostGroup: #{host_group} in Cluster: #{name}" if result_host_group.nil?
+        result_host_group
+      end
+
+      # Caches and returns the list of hosts under a cluster resource. It might
+      # be different from total hosts in a cluster if a host group is specified.
+      #
+      # If a host group is present,  only the hosts under the host group are
+      # considered.
+      #
+      # @return [List] List of hosts in the cluster resource.
+      def host
+        return @host if @host
+        @host = host_group_mob.nil? ? mob.host : host_group_mob.host
       end
 
       # @return [String] cluster name.
@@ -80,10 +137,46 @@ module VSphereCloud
         accessible_datastores.select { |name, datastore| name =~ pattern }
       end
 
+      # Fetches memory utilization of host group. calculates raw available memory.
+      #
+      # Raises an error if and only if all hosts fail to report stats.
+      #
+      # @return RAW available memory in the host group
+      def fetch_host_group_utilization
+        # Reject all non-normal state hosts.
+        # Maintenance hosts report all stats so we need to filter them
+        logger.debug("Fetching Memory utilization for Host Group #{host_group}")
+        healthy_hosts = host_group_mob.host.select do |host|
+          host.runtime.connection_state == 'connected' && !host.runtime.in_maintenance_mode
+        end
+        if healthy_hosts.empty?
+          logger.debug("No fit host found in #{host_group}") if healthy_hosts.empty?
+          return 0
+        end
+
+        total_memory = healthy_hosts.map do |host|
+          host.hardware&.memory_size
+        end.compact.sum
+
+        # Return immediately if hosts in host group fail to report memory stats.
+        if total_memory == 0
+          logger.debug("Host Group #{host_group} reported total memory statistics as 0")
+          return 0
+        end
+
+        used_memory = healthy_hosts.map do |host|
+          # Overall memory usage is in MB. Convert it to Bytes.
+          host.summary.quick_stats.overall_memory_usage * BYTES_IN_MB
+        end.compact.sum
+        logger.warning("Host Group #{host_group} reported 0 bytes used memory") if used_memory == 0
+        (total_memory - used_memory) / BYTES_IN_MB
+      end
+
       # Fetches the raw cluster utilization from vSphere.
       #
       # @return
       def fetch_cluster_utilization()
+        logger.debug("Fetching Memory utilization for Cluster #{self.mob.name}")
         properties = @client.cloud_searcher.get_properties(mob, Vim::ClusterComputeResource, 'summary')
         raise "Failed to get utilization for cluster'#{self.mob.name}'" if properties.nil?
 
@@ -101,6 +194,7 @@ module VSphereCloud
       #
       # @return [void]
       def fetch_resource_pool_utilization
+        logger.debug("Fetching Memory utilization for Resource Pool #{resource_pool.name}")
         properties = @client.cloud_searcher.get_properties(resource_pool.mob, Vim::ResourcePool, 'summary')
         raise "Failed to get utilization for resource pool '#{resource_pool}'" if properties.nil?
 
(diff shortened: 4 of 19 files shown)


src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
```diff
@@ -32,6 +32,23 @@ def add_vm(vm)
       end
     end
 
+    # Adds VM to given VM Group & creates VM/HOST Affinity rule
+    # @param [Vim::VirtualMachine] vm
+    # @params [String] vm_group_name
```
ktchen14: I think this should be `@param`.

src/vsphere_cpi/lib/cloud/vsphere/cluster_config.rb
```diff
@@ -8,8 +8,8 @@ def initialize(name, config_hash)
       @config = config_hash
     end
 
-    def resource_pool
-      @config['resource_pool']
+    %w[vm_group host_group vm_host_affinity_rule_name resource_pool].each do |name|
```
ktchen14: This is terse but to be honest given how short this file is already is doing just:

```ruby
def vm_group; @config['vm_group']; end
def host_group; @config['host_group']; end
...
```

Is better in this case in my mind.

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/vm_group.rb
```diff
@@ -59,8 +59,37 @@ def delete_vm_groups(vm_group_names)
...
+        # Now, there is a potential problem here if anyone creates a rule
+        # outside of CPI using this vm_group, it will be deleted too. To narrow
+        # the scope of this potential bug, CPI will only delete the
+        # vm-host-group-affinity rule type. This is because, currently it only
+        # creates vm-host-affinity rule using vm groups.
+        #
+        # Long Term Solution is to disallow admin to specify their own rule
+        # names and CPI generates a rule name from vm_group and hence will be
+        # able to identify correct rule to be deleted.
+        rules_spec = empty_vm_groups.reduce([]) do |rules_spec, vm_group|
```
ktchen14: In general `reduce` in Ruby is discouraged. I'm not exactly sure but I think in this case we can use `each_with_object` instead.

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
```diff
@@ -45,12 +46,72 @@ def free_memory
...
+      # Caches and returns the list of hosts under a cluster resource. It might
+      # be different from total hosts in a cluster if a host group is specified.
+      #
+      # In case a host group is present only the hosts under the host group are
+      # considered.
+      #
+      # If the host group does not exist/wrong it returns all the hosts
+      #
+      # @return [List] List of hosts in the cluster resource.
+      def host
```
nehagjain15: should be names hosts since it is returning a list of hosts and not a single host

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
```diff
@@ -458,6 +458,30 @@ def find_child_by_name(mob, child_path)
       find_child_by_name(mob.child_entity.find {|c| c.name == child_entity_name }, child_path)
     end
 
+    def clean_up_vm_groups(cluster, vm_group_names)
```
nehagjain15: Before merging should refactor this so that these methods are not duplicated in vm_group and here.

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
EleanorRigby: I agree. I will refactor this in another PR.  This was just kept host because of extensive usage and I did not want to change so many unrelated files.

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
```diff
@@ -59,6 +75,30 @@ def update_rule(rule_key)
       add_anti_affinity_rule(VimSdk::Vim::Option::ArrayUpdateSpec::Operation::EDIT, rule_key)
     end
 
+    def find_vm_group(vm_group_name)
```
EleanorRigby: Don't think this is getting called from anywhere. Can delete it

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/vm_group.rb
```diff
@@ -59,15 +59,53 @@ def delete_vm_groups(vm_group_names)
           group_spec.remove_key = vm_group.name
           group_spec
         end
+
+        # If these vm_groups are getting deleted. Also delete all the
+        # vm-host group rules associated with it.
+        #
+        # @TODO : CAVEAT: if anyone creates a rule outside of CPI using these
+        # vm_groupS, IT will be deleted too. To narrow
```
EleanorRigby: typo in comment

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/vm_group.rb
```diff
@@ -59,15 +59,53 @@ def delete_vm_groups(vm_group_names)
...
+        # the scope of this caveat, CPI will only delete the
+        # vm-host-group-affinity rule type. This is because, currently it only
+        # creates vm-host-affinity rule using vm groups.
+        #
+        # Long Term Solution is to disallow admin to specify their own rule
```
EleanorRigby: remove this part of comment

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
nehagjain15: This method can be deleted as it is not being called from anywhere

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -114,6 +114,36 @@ def create(vm_config)
           raise "Unable to parse vmx options: 'vmx_options' is not a Hash"
         end
 
+        host = nil
+        # Before cloning we need to make sure the host is correctly picked up
+        # from cluster's host group if provided.
+        unless cluster.host_group.nil?
```
nehagjain15: this can be just if cluster.host_group

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -114,6 +114,36 @@ def create(vm_config)
...
+          # placement_spec = VimSdk::Vim::Cluster::PlacementSpec.new
```
nehagjain15: since, this Since, this is for future implementation, should remove it from here before we merge to master

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -114,6 +114,36 @@ def create(vm_config)
...
+          #   logger.warning("received #{e} while asking DRS for recommendation")
+          #   host = cluster.host_group_mob.host.find do |host|
+          #     host.runtime.connection_state == 'connected' &&
+          #       !host.runtime.in_maintenance_mode
+          #   end
+          # end
+          # @TA:TODO : Stop-Gap measure until DRS team solves bug for placeVM below.
+          # Selecting a random host out of all host groups to get initial placement correct
+          # Filtering it for healthy host.
+          host = cluster.host_group_mob.host.find do |host|
```
nehagjain15: if there is no healthy host, should you error out here?

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -139,19 +170,27 @@ def create(vm_config)
...
 
           # DRS Rules
           create_drs_rules(vm_config, created_vm.mob, cluster)
 
-          # Add vm to VMGroup
-          add_vm_to_vm_group(vm_config, created_vm.mob, cluster)
+          # Add vm to VM Group present in vm_type
+          add_vm_to_vm_group(vm_config.vm_type.vm_group, created_vm_mob, cluster)
+
+          unless cluster.host_group.nil?
```
nehagjain15: if instead of unless

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -208,13 +247,24 @@ def create_drs_rules(vm_config, vm_mob, cluster)
...
+      return if vm_group.nil?
+      vm_group_creator = VSphereCloud::VmGroup.new(
+        @client,
+        cluster.mob
+      )
+      vm_group_creator.add_vm_to_vm_group(vm_mob, vm_group)
+    end
+
+    def create_vm_host_affinity_rule(vm_mob, cluster)
+      return if cluster.host_group.nil?
```
nehagjain15: This check is already happening before the method is being called, either you can remove this, or remove the check when this method is called as, add_vm_to_vm_group method also has a check if vm_group.nil?

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_lock.rb
```diff
@@ -29,6 +29,8 @@ def with_drs_lock
 
     private
 
+    attr_reader :drs_lock_name
```
nehagjain15: Not needed, can be removed

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/drs_lock_spec.rb
```diff
@@ -2,7 +2,8 @@
 require 'timecop'
 
 describe VSphereCloud::DrsLock, fake_logger: true do
-  subject(:drs_lock) { described_class.new(vm_attribute_manager) }
+  subject(:drs_lock) { described_class.new(vm_attribute_manager, drs_lock_name) }
```
nehagjain15: This is using the default lock_name so you dont need to pass it, in end there is a test to check a custom drs_lock name

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/drs_lock_spec.rb
```diff
@@ -2,7 +2,8 @@
...
+  let(:drs_lock_name) {'drs_lock'}
```
nehagjain15: This can be removed refer comment before this

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/drs_rule_spec.rb
```diff
@@ -151,4 +151,48 @@ def self.it_adds_custom_attribute_to_vm
...
+      it 'returns the rule with DRS lock' do
+        expect(drs_lock).to receive(:with_drs_lock).and_yield.ordered
+        expect(datacenter_cluster).to_not receive(:reconfigure_ex)
+        expect(drs_rule).to_not receive(:create_vm_host_affinity_rule)
+        drs_rule.add_vm_host_affinity_rule(vm, vm_group_name, host_group_name)
+      end
+    end
+    context 'when rule does not exist' do
+      let(:rules) { [] }
+      it 'adds vm to the vm_group and creates DRS rule with DRS lock' do
```
nehagjain15: reword, as this method only creates vm_host_afffinity_rule and does not add vm to vm_group

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/drs_rule_spec.rb
```diff
@@ -151,4 +151,48 @@ def self.it_adds_custom_attribute_to_vm
     end
   end
 
+  describe '#add_vm_host_affinity_rule' do
+    let(:vm) { instance_double('VimSdk::Vim::VirtualMachine', name: 'fake-vm') }
+    let(:fake_vm_1) { instance_double('VimSdk::Vim::VirtualMachine', name: 'fake-vm-1') }
+    let(:vm_group_name) {'fake-vm-group'}
+    let(:host_group_name) {'fake-host-group'}
+    let(:configuration_ex) { double(:configuration_ex, rule: rules, group: vm_groups) }
+    let(:vm_groups) { [ VimSdk::Vim::Cluster::VmGroup.new(name: vm_group_name, vm: [fake_vm_1]) ] }
```
nehagjain15: this  cam be removed as this method doesn't use the vm_group

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/drs_rule_spec.rb
```diff
@@ -151,4 +151,48 @@ def self.it_adds_custom_attribute_to_vm
     end
   end
 
+  describe '#add_vm_host_affinity_rule' do
+    let(:vm) { instance_double('VimSdk::Vim::VirtualMachine', name: 'fake-vm') }
+    let(:fake_vm_1) { instance_double('VimSdk::Vim::VirtualMachine', name: 'fake-vm-1') }
+    let(:vm_group_name) {'fake-vm-group'}
+    let(:host_group_name) {'fake-host-group'}
+    let(:configuration_ex) { double(:configuration_ex, rule: rules, group: vm_groups) }
```
nehagjain15: you can remove the group from here as it is not being used

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
```diff
@@ -32,6 +32,22 @@ def add_vm(vm)
       end
     end
 
+    # Adds VM to given VM Group & creates VM/HOST Affinity rule
+    # @param [Vim::VirtualMachine] vm
+    # @param [String] vm_group_name
+    # @param [String] host_group_name
+    def add_vm_host_affinity_rule(vm, vm_group_name, host_group_name)
```
nehagjain15: vm is not needed with current implementation. I think we can refactor later to make addition of vm to vm_group as part of this code rather then doing it separately

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/drs_rule_spec.rb
```diff
@@ -151,4 +151,48 @@ def self.it_adds_custom_attribute_to_vm
     end
   end
 
+  describe '#add_vm_host_affinity_rule' do
+    let(:vm) { instance_double('VimSdk::Vim::VirtualMachine', name: 'fake-vm') }
+    let(:fake_vm_1) { instance_double('VimSdk::Vim::VirtualMachine', name: 'fake-vm-1') }
```
nehagjain15: vm, fake_vm_1 can all be removed once the add_vm_host_affinity_rule definition is updated

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/vm_group_spec.rb
```diff
@@ -107,4 +110,47 @@ def with_lock
...
     end
   end
+
+  describe '#get_delete_rule_spec' do
+    before do
+      allow(datacenter_cluster).to receive(:configuration_ex).and_return(configuration_ex)
+    end
+    context 'when no rule exists' do
+      before do
+        allow(configuration_ex).to receive(:rule).and_return(nil)
```
nehagjain15: I think this returns an empty array, should confirm the with an actual call and setup the test accordingly

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/vm_group.rb
```diff
@@ -59,15 +59,53 @@ def delete_vm_groups(vm_group_names)
...
         end
       end
     end
 
     private
+    def get_delete_rule_spec(empty_vm_groups)
+      if @cluster.configuration_ex.rule.nil?
+        return nil
+      end
+      rules_spec= empty_vm_groups&.each_with_object([]) do |vm_group, rules_spec|
```
nehagjain15: dont need an & check as empty_vm_groups is an array and the check is being made when this is set

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/vm_group.rb
```diff
@@ -59,15 +59,53 @@ def delete_vm_groups(vm_group_names)
...
         @client.wait_for_task do
           @cluster.reconfigure_ex(config_spec, true)
         end
       end
     end
 
     private
+    def get_delete_rule_spec(empty_vm_groups)
+      if @cluster.configuration_ex.rule.nil?
+        return nil
```
nehagjain15: best would be this returns an empty array so return type is same and then called can check if spec is empty or not

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/vm_group_spec.rb
```diff
@@ -107,4 +110,47 @@ def with_lock
...
+      end
+      it 'returns nil' do
+        expect(vm_group.send(:get_delete_rule_spec, *[vm_groups])).to be(nil)
```
nehagjain15: as I commented should return empty array

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/vm_group_spec.rb
```diff
@@ -107,4 +110,47 @@ def with_lock
...
+      end
+    end
+    context 'when vm_groups passed are empty' do
+      before do
+        allow(configuration_ex).to receive(:rule).and_return(['fake-rule'])
+      end
+      it 'returns nil' do
+        expect(vm_group.send(:get_delete_rule_spec, *[nil])).to be(nil)
```
nehagjain15: spec says empty vm_groups but you are passing a nil here

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/vm_group_spec.rb
```diff
@@ -107,4 +110,47 @@ def with_lock
...
+      end
+    end
+    context 'when vm groups passed are not empty' do
```
nehagjain15: reword "when vm_groups are present"

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_config_spec.rb
```diff
@@ -266,7 +266,8 @@ def fake_datastore(name, free_space, mob = nil)
             accessible_datastores: {
               'smaller-ds'=> smaller_ds,
             },
-            mob: cluster1_mob
+            mob: cluster1_mob,
+            host: [host_system],
```
nehagjain15: then host can be removed from cluster1_mob and cluster2_mob definition as it is not being used

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/cluster_spec.rb
```diff
@@ -262,6 +375,116 @@ def generate_host_property(mob:, name:, connection_state:, maintenance_mode:, me
...
+          it 'returns nil' do
+            expect(cluster.host_group).to be(nil)
+          end
+        end
+        context 'when host group is defined in config' do
+          let(:cluster_config) do
+            instance_double(
+              'VSphereCloud::ClusterConfig',
+              name: 'fake-cluster-name',
+              resource_pool: 'fake-resource-pool',
```
nehagjain15: resource_pool can be removed

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/cluster_spec.rb
```diff
@@ -262,6 +375,116 @@ def generate_host_property(mob:, name:, connection_state:, maintenance_mode:, me
...
+        it 'returns host_group name suffixed with _vm_group' do
+          expect(cluster.vm_group).to eq(cluster.host_group+VSphereCloud::Resources::Cluster::CLUSTER_VM_GROUP_SUFFIX)
+        end
+      end
+      context 'when host group name is longer than 100 characters' do
+        let(:cluster_config) do
+          instance_double(
+            'VSphereCloud::ClusterConfig',
+            name: 'fake-cluster-name',
+            resource_pool: 'fake-resource-pool',
```
nehagjain15: resource_pool can be removed

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/cluster_spec.rb
```diff
@@ -262,6 +375,116 @@ def generate_host_property(mob:, name:, connection_state:, maintenance_mode:, me
       end
     end
 
+    describe 'host groups and vm groups are defined in AZ' do
+      describe '#vm_group' do
+        let(:cluster_config) do
+          instance_double(
+            'VSphereCloud::ClusterConfig',
+            name: 'fake-cluster-name',
+            resource_pool: 'fake-resource-pool',
```
nehagjain15: resource_pool can be removed

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/cluster_spec.rb
```diff
@@ -262,6 +375,116 @@ def generate_host_property(mob:, name:, connection_state:, maintenance_mode:, me
...
+          it 'returns nil' do
+            expect(cluster.vm_host_affinity_rule_name).to be(nil)
+          end
+        end
+        context 'when host group is defined' do
+          let(:cluster_config) do
+            instance_double(
+              'VSphereCloud::ClusterConfig',
+              name: 'fake-cluster-name',
+              resource_pool: 'fake-resource-pool',
```
nehagjain15: resource_pool can be removed and similarly below from all references

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/cluster_spec.rb
```diff
@@ -262,6 +375,116 @@ def generate_host_property(mob:, name:, connection_state:, maintenance_mode:, me
...
+              host_group: 'fake-host-group',
```
nehagjain15: you could just use let(:host_group) in all the tests instead of redefining cluster_config with host_group.

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
```diff
@@ -0,0 +1,402 @@
...
+        it 'should create one VM on the second cluster without host group\
+            (as resource pool has more memory available than host group)' do
+          simple_vm_lifecycle(cpi, '', vm_type, get_network_spec) do |vm_id|
+            vm = cpi.vm_provider.find(vm_id)
+            expect(vm.cluster).to eq(@second_cluster_name)
+          end
+          expect(get_count_vm_host_affinity_rules(cluster_mob) +
+            get_count_vm_host_affinity_rules(second_cluster_mob)).to eq(0)
+        end
+        it 'should create multiple VMs (3) one after another on the second cluster\'s resource pool' do
```
nehagjain15: creation of multiple VM's for different configurations is repetitive and I think unnecessary for all the context in this spec file

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
```diff
@@ -0,0 +1,402 @@
...
+
+    context 'when there is only one vSphere cluster in an AZ' do
+      let(:hosts_in_first_host_group) { first_host_group_mob.host.map(&:name) }
+      let(:hosts_in_second_host_group) { second_host_group_mob.host.map(&:name) }
+
+      context 'when one host_group is specified in the cluster' do
+        it 'should create one VM on one of the hosts in the host groups' do
+          simple_vm_lifecycle(cpi, '', vm_type, get_network_spec) do |vm_id|
+            vm = cpi.vm_provider.find(vm_id)
+            expect(hosts_in_first_host_group).to include(vm.mob.runtime.host.name)
```
nehagjain15: you should also test as part of this integration test that VM is added to the VM group and the rule is present. This spec is no where testing anything about VM Host Rule creation except that the count is 0 at the end

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
EleanorRigby: I would prefer keeping it. It is small and is not making things messy. Will be super easy to uncomment and push later.

Plus, this PR might uncomment it later depending on test runs with and without this change today.

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
EleanorRigby: `vm` is removed. I am not sure about merging addition of vm group to this function or I do not understand your intent. What  value would it bring rather than doing this way?

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
EleanorRigby: I would prefer explicit nil? everywhere as it differentiates `false` from `nil` .  makes it clear that host_group is not a boolean. It does not sees so from the name but practice has to be consistent across variables.

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
EleanorRigby: I had a TODO here before with title `should we make a function secure or the callee place should make sure of this` 

Since this is a local cluster resource and check will happen very fast, I decided to keep it here also in case if this function is used in future.
What do you say?

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/vm_group_spec.rb
EleanorRigby: added one more test to check empty. letting nil stay.

src/vsphere_cpi/spec/unit/cloud/vsphere/drs_rules/vm_group_spec.rb
EleanorRigby: Corrected.

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/cluster_spec.rb
EleanorRigby: I did a bit of refactor. It would eliminate some redundancy but not much. i will keep it same for now.

src/vsphere_cpi/spec/unit/cloud/vsphere/vm_config_spec.rb
EleanorRigby: It is being used. Caused error when i removed.

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
EleanorRigby: Letting this stay in favor of repeated testing for any non-deterministic bug.
Will take it our after the release is cut and feature stabilizes.

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
EleanorRigby: On it.

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
EleanorRigby: Done. Fixed. Good catch

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
EleanorRigby: To be discussed later for future releases

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
nehagjain15: Have created a follow up story to resolve this.

Issue #174, opened 2018-11-23
Datastore usage appears unbalanced after recent upgrades


thelangley:
**The (potential) Bug**
We've recently moved to vsphere-cpi v51.0.2 from v50 and upgraded our vCenters to 6.5 u2c

We're finding that when a new stemcell is being deployed, it has a tendency to have a "favourite" Datastore so much so that it tries to almost fill it to 100%, even if loads of the others have a fair amount free.

```
- type: replace
  path: /instance_groups/name=bosh/properties/vcenter?
  value: &vcenter
    address: ((fqdn))
    user: ((user))
    password: ((theword))
    datacenters:
    - name: ((my_dc_name))
      vm_folder: ((my_folder))
      template_folder: ((my_template_folder))
      datastore_pattern: ((my_datastore_pattern))
      persistent_datastore_pattern: ((my_persistent_disk_datastore_pattern))
      disk_path: ((my_amazing_path))
      clusters:
      - ((vsphere_cluster)): { ((poolies_info)) }
```

**To Reproduce**
1. Ensure you have many datastores in vCenter available for action
2. Define them as a regex and use the pattern in your ops file
3. Deploy
4. Recreate the things
5. Recreate the things

Over time you will see one datastore gets more and more full

**CPI Error Log**
This is no error in the logs but I can see the decisions being made.

It's using the regex nicely to find all the Datastores which match but then the roll of the dice usually selects a particular one.

```
D, [2018-11-23T12:29:16.747104 #31769] DEBUG -- [req_id cpi-212057]: All clusters provided: {"my_amazing_cluster"=>#<VSphereCloud::ClusterConfig:xxxxxx @name="my_amazing_cluster", @config={"resource_pool"=>"my_amazing_resource_pool"}>}
D, [2018-11-23T12:29:16.747207 #31769] DEBUG -- [req_id cpi-212057]: Running method 'FindByInventoryPath'...
I, [2018-11-23T12:29:16.783249 #31769]  INFO -- [req_id cpi-212057]: Using global ephemeral disk datastore pattern: my_amazing_datastore_[0-9]{2}
[...]

PR #175, opened 2018-11-26
Fix Skew in datastore selection

EleanorRigby:
# Description

Datastore with most free space should be scored first in the list of placemnts emitted out by disk placement selection pipeline

## Related PR and Issues
Fixes Issue #174 

Datastore Selection

## Type of change

## How Has This Been Tested?
 Unit and Integration test Added.


src/vsphere_cpi/spec/integration/cluster_different_datastores_spec.rb
```diff
@@ -33,72 +33,123 @@
...
+      options = cpi_options(
+        datacenters: [{
+          datastore_pattern: '.*',
+          persistent_datastore_pattern: '.*',
+        }],
+      )
+      VSphereCloud::Cloud.new(options)
+    end
+
+    it 'should pick the datastore with most amount of free space' do
```
nehagjain15: I think this test should be part of a new set of integration tests eg datastore_selection_spec.rb. So, that we can more integration tests related to datastore selection over there and this file will remain a test for testing using different cpis

src/vsphere_cpi/spec/integration/cluster_different_datastores_spec.rb
```diff
@@ -33,72 +33,123 @@
...
+      cluster_resource = all_datastore_cpi.datacenter.find_cluster(@cluster_name)
+      max_free_space_datastore = cluster_resource.mob.datastore.sort_by do |ds|
```
nehagjain15: Can this be not a specific datastore ? If according to our env setup if you expect it is datastore-x then lets use that instead of dynamic value for datastore

EleanorRigby: This will be `nsf0-1` but in some cases if there is lot of stemcell replication or repeated BATS test, `vsandatstore` might trump `nfs0-1` in free space. Therefore, it is not possible to keep it static.

src/vsphere_cpi/spec/integration/cluster_different_datastores_spec.rb
EleanorRigby: The last heading of file before I added my changes was `given cpis that are configured to use same cluster but different datastores`
That is exactly what I solved. We can instead rename this file to what you suggested.

PR #180, opened 2018-12-06
De-duplicate NSX security groups

EleanorRigby:
# Description
Security groups for NSX-V has multiple sources
 - vm_type
 - bosh groups in create_vm env hash
 - nsx lbs

We need to gather all `nsx-security-groups` and deduplicate before calling API to add vm to these `nsx-security-groups`. It saves execution time. More robust code as we do not need to rely on some fragile logic to  not raise error if return code is 203 (`vm` already exists in Security group). 

The issue is caused when upgrading from NSX-V 6.3.x to 6.4.2. 
6.4.2 NSX manager's security group API return code is different from 6.3.x

## Related PR and Issues
Fixes issue `Object vm-32660 is already a member of Security5.`

Adding NSX-Security groups

## Type of change

- [x]  Performance Improvement

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -347,23 +347,40 @@ module VSphereCloud
         end
 
         begin
-          vm_type.nsx_security_groups.each do |security_group|
-            nsx.add_vm_to_security_group(security_group, created_vm.mob_id)
-          end unless vm_type.nsx_security_groups.nil?
-
 
           if @config.nsx_enabled?
-            bosh_groups = (environment || {}).fetch('bosh', {}).fetch('groups', [])
-            bosh_groups.each do |security_group|
+            all_nsx_security_groups = Set.new
+
+            # Gather all security groups from three sources into a Set
+            # 1. VM Type
+            # 2. BOSH Groups
+            # 3. NSX Load Balancers
+
+            # VM Type
+            vm_type_security_group = vm_type.nsx_security_groups
+            all_nsx_security_groups.merge(vm_type_security_group) unless vm_type_security_group.nil?
+
+            # BOSH Group
+            bosh_groups_security_groups = (environment || {}).fetch('bosh', {}).fetch('groups', [])
+            all_nsx_security_groups.merge(bosh_groups_security_groups) unless bosh_groups_security_groups.nil?
+
+            # Load Balancer
+            unless vm_type.nsx_lbs.nil?
+              nsx_lbs_security_groups = vm_type.nsx_lbs.map { |m| m['security_group'] }
+              all_nsx_security_groups.merge(nsx_lbs_security_groups.compact)
+            end
+
+            # Add vm to security groups
+            all_nsx_security_groups.each do |security_group|
               nsx.add_vm_to_security_group(security_group, created_vm.mob_id)
+            end
+
+            # Add vm to load balancer
+            unless vm_type.nsx_lbs.nil? || vm_type.nsx_lbs.empty?
+              nsx.add_members_to_lbs(vm_type.nsx_lbs)
             end
           end
 
-          unless vm_type.nsx_lbs.nil?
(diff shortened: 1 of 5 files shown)


src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -347,23 +347,40 @@ def create_vm(agent_id, stemcell_cid, vm_type, networks_spec, existing_disk_cids
...
+            all_nsx_security_groups.merge(bosh_groups_security_groups) unless bosh_groups_security_groups.nil?
+
+            # Load Balancer
+            unless vm_type.nsx_lbs.nil?
+              nsx_lbs_security_groups = vm_type.nsx_lbs.map { |m| m['security_group'] }
+              all_nsx_security_groups.merge(nsx_lbs_security_groups)
+            end
+
+            # Add vm to security groups
+            all_nsx_security_groups.each do |security_group|
```
nehagjain15: Please add unit test for this

EleanorRigby: I am on it. Just wanted to make this an incremental PR.

EleanorRigby:
@nehagjain15 : Can you review it once more?

src/vsphere_cpi/spec/integration/nsx_vsphere_spec.rb
```diff
@@ -177,7 +177,7 @@
         cpi_options
       end
 
-      it 'raises an error' do
+      xit 'raises an error' do
```
nehagjain15: why is this commented?

nehagjain15: I think you can delete it as this test is no longer valid with your refactoring in cloud.rb

src/vsphere_cpi/spec/unit/cloud/vsphere/cloud_spec.rb
```diff
@@ -785,6 +788,267 @@ module VSphereCloud
...
+            allow(VmConfig).to receive(:new).and_return(vm_config)
+            allow(NsxHttpClient).to receive(:new)
+              .with('fake-nsx-user', 'fake-nsx-password', 'fake-log-file')
+              .and_return(http_basic_auth_client)
+            allow(NSX).to receive(:new).and_return(nsx)
+
+            expect(VmCreator).to receive(:new).and_return(vm_creator)
+            expect(vm_creator).to receive(:create).with(vm_config).and_return(fake_vm)
+            expect(cloud_config).to receive(:validate_nsx_options)
+            expect(nsx).to receive(:add_vm_to_security_group).exactly(3).times
```
nehagjain15: you could also add a check expect(nsx).to receive(:add_vm_to_security_group).with(/(fake_duplicate_sg| another_fake_duplicate_sg|fake_unique_bosh_group)/, fake_vm.mob_id)

src/vsphere_cpi/spec/unit/cloud/vsphere/cloud_spec.rb
```diff
@@ -785,6 +788,267 @@ module VSphereCloud
...
```
nehagjain15: except this line almost all tests have same expectations, best is to refactor so that every 'it' block doesn't have any duplication.
Similarly for vm_type you could extract it out and each context can just set security_groups and lbs

src/vsphere_cpi/spec/unit/cloud/vsphere/cloud_spec.rb
EleanorRigby: Yeah. that would be nice. I will add it to this commit but not for the already back ported once.

src/vsphere_cpi/spec/unit/cloud/vsphere/cloud_spec.rb
EleanorRigby: Can I take common expectation out and put them in a block? How?

If I extract vm_type out , I am only extracting ram, cpu out. Other things will still stay, so letting it be.

PR #187, opened 2019-01-11
Switch should be removed even if there were zero operational ports

kreamyx:
Fixes the integration test error


EleanorRigby:
Do we have tests which check all cases 0, 1, and > 1?

PR #189, opened 2019-01-18
[Test in Progress] Folder up disks while attaching them

EleanorRigby:
- Objective of the change is to prevent SDRS from running into file
conflicts while relocating VM.
- Unfolder disk to bosh disk folder when detaching them
 - Remove has_persistent_disk_property_mismatch function for no usage
 - move_persistent_disk_to_datastore function now takes a  VM CID as
 required argument. Since it is acalled from only one place the change
 should be good.


EleanorRigby:
@ktchen14 : Shouldn't it be `\b` where \b matches word boundary

EleanorRigby:
Closing this as solution unlikely to prevent disk loss in event of CPI process crash.

PR #190, opened 2019-01-25
[POC][Expermental] Inventorization of VMs by deployment name.

EleanorRigby:
# Description

This change creates deployment named folder in vSphere Inventory for each BOSH Deployment

Steps:

Create_VM
1. Capture bosh deployment name from environment groups passed in create vm
2. Take new lock
3. Create Folder
4. Clone VM
5. Add VM Folder Path in Custom Attr
5. Release Lock

Delete_VM
1. Get VM Folder Path from custom attr, Extract vm parent folder
2. Delete VM
3. Get lock (same as in create)
4. Delete folder if no vm present.
5. Release Lock

### Disadvantage
Mutex is needed to cleanly delete bosh deployment named inventory folder. As VMs are constantly getting created and deleted, CPI needs synchronization to be able to decide when to delete deployment folder.
As a result, cloning and delete operations in vSphere becomes single threaded for one deployment.

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere.rb
@@ -71,8 +71,8 @@ require 'cloud/vsphere/vm_placement_selection_pipeline'
 
 
 module Bosh
-  module Clouds
-    class VSphere
+    module Clouds
+      class VSphere
       extend Forwardable
 
       def_delegators :@delegate,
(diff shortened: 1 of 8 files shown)


EleanorRigby:
Closing this , pending discussion on lifecycle management of deployment named inventory folders. 

----
#### Details
Currently, CPI has no way of knowing when a deployment is getting deleted. This is essential to know if CPI needs to delete the inventory folder based on deployment names. 
One solution is to serialize creation and deletion of VMs , which will seriously affect performance.

PR #192, opened 2019-02-12
Add support for nested resource pool specification.

### LINKED ISSUE #11: Could not find 'VimSdk::Vim::ResourcePool
take from https://github.com/cloudfoundry/cf-mysql-release/issues/134

Could anyone help me on this please?

```
Started compiling packages
Started compiling packages > cluster-schema-verifier/a2f8659250e5b51a7f5d0b0fdff5dbdc5ee98895
Started compiling packages > cli/41edd6a70de171fcefd9328c5a616a03e9c88677
Started compiling packages > mysqlclient/ce95f8ac566f76b650992987d5282ee473356e43
Started compiling packages > ruby/a57562f234ecf56581a73f2e18715f19ea74e9f1
Failed compiling packages > mysqlclient/ce95f8ac566f76b650992987d5282ee473356e43: Unknown CPI error 'Unknown' with message 'Could not find 'VimSdk::Vim::ResourcePool': {:root=>
#<VimSdk::Vim::ResourcePool:0x005559933b4380
```

EleanorRigby:
- Adds support for specifying a resource pool path under a cluster.
- Currently, only a single name is accepted for resource pool and in
cases where multiple nested resource pools have same names, CPI fails to
identify them correctly; often returning the first match.
- This change introduces finding resource pools by inventory path. It is
a supported method in vCenter's search index object and is expected to
be much faster than propoerty collector.
- The fully qualified name of a resource pool is expected to be of form
  "Datacenter/host/Cluster/Resources/<Resource-Pool-Path (Relative to
  Cluster)"
  The CPI expects user to provide just the "<Resource-Pool-Path
  (Relative to Cluster)"

[#163777434](https://www.pivotaltracker.com/story/show/163777434)

# Description

Allows admin/operator to specify resource pool path (relative to the parent cluster)

## Related PR and Issues
Fixes #11 

Should Improve resource pool search performance.

## Type of change

## How Has This Been Tested?
- [x] New & Existing Unit Tests
- [x] New & Existing Integration tests

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/resource_pool.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/resource_pool.rb
@@ -11,10 +11,11 @@ module VSphereCloud
       # @param [Cluster] cluster parent cluster.
       # @param [Vim::ResourcePool] root_resource_pool cluster's root resource
       #   pool.
-      def initialize(client, cluster_config, root_resource_pool)
+      def initialize(client, cluster_config, root_resource_pool, datacenter_name)
         @cluster_config = cluster_config
         @root_resource_pool = root_resource_pool
         @client = client
+        @datacenter_name = datacenter_name
       end
 
       def name
@@ -27,12 +28,24 @@ module VSphereCloud
         if @cluster_config.resource_pool.nil?
           @mob = @root_resource_pool
         else
-          client = @client
-          @mob = client.cloud_searcher.get_managed_object(
-            Vim::ResourcePool,
-            :root => @root_resource_pool,
-            :name => @cluster_config.resource_pool)
-          logger.debug("Found requested resource pool: #{@mob}")
+          begin
+            resource_pool_path_suffix = @cluster_config.resource_pool
+            datacenter_cluster_path_prefix = "#{@datacenter_name}/host/#{@cluster_config.name}/Resources/"
+            full_inventory_path = datacenter_cluster_path_prefix + resource_pool_path_suffix
+            # Replace all multiple consecutive / with single /
+            full_inventory_path.gsub!(/\/+/, '/').chomp!('/')
+            @mob = @client.service_content.search_index.find_by_inventory_path(full_inventory_path)
+            raise "Could not find resource pool #{resource_pool_path_suffix} for inventory path #{full_inventory_path}" if @mob.nil?
+          rescue => e
+            logger.info("#{e} - #{e.backtrace.join("\n")}")
+            logger.info("Trying to find #{@cluster_config.resource_pool} through property collector")
+            @mob = @client.cloud_searcher.get_managed_object(
+                Vim::ResourcePool,
+                :root => @root_resource_pool,
+                :name => @cluster_config.resource_pool)
+            raise "Could not find resource pool #{@cluster_config.resource_pool}" if @mob.nil?
+            logger.debug("Found requested resource pool: #{@mob}")
+          end
         end
         @mob
       end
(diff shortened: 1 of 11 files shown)


src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -125,9 +125,10 @@ def has_custom_cluster_properties?
     
     def find_clusters(clusters_spec)
       clusters = []
+      datacenter_name = vm_type.datacenter.name
```
ktchen14: Can we make this entire method just use `map` instead of this whole non-Rubyish pattern?

src/vsphere_cpi/lib/cloud/vsphere/resources/resource_pool.rb
```diff
@@ -27,12 +28,24 @@ def mob
...
         else
-          client = @client
-          @mob = client.cloud_searcher.get_managed_object(
-            Vim::ResourcePool,
-            :root => @root_resource_pool,
-            :name => @cluster_config.resource_pool)
-          logger.debug("Found requested resource pool: #{@mob}")
+          begin
+            resource_pool_path_suffix = @cluster_config.resource_pool
+            datacenter_cluster_path_prefix = "#{@datacenter_name}/host/#{@cluster_config.name}/Resources/"
```
ktchen14: This is dangerous if there are special characters in `@datacenter_name` or `@cluster_config.name`. I believe that you can make this safe by URL escaping those strings.

src/vsphere_cpi/lib/cloud/vsphere/resources/resource_pool.rb
```diff
@@ -27,12 +28,24 @@ def mob
...
+            full_inventory_path = datacenter_cluster_path_prefix + resource_pool_path_suffix
+            # Replace all multiple consecutive / with single /
+            full_inventory_path.gsub!(/\/+/, '/').chomp!('/')
```
ktchen14: In which case this may be unnecessary.

src/vsphere_cpi/lib/cloud/vsphere/resources/resource_pool.rb
```diff
@@ -27,12 +28,24 @@ def mob
...
+            @mob = @client.service_content.search_index.find_by_inventory_path(full_inventory_path)
+            raise "Could not find resource pool #{resource_pool_path_suffix} for inventory path #{full_inventory_path}" if @mob.nil?
+          rescue => e
```
ktchen14: I'm not sure I like this pattern:

```ruby
begin
  raise ...
rescue => e
  raise
end
```

I think a simple `if` statement with an early return when found is more appropriate here.

src/vsphere_cpi/lib/cloud/vsphere/resources/resource_pool.rb
EleanorRigby: I don't think escape will remove unnecessary `/` characters.
My concern basically is that the user while defining a resource pool can do it in these ways 
 - `vcpi-rp-1-1/vcpi-sub-rp`
 - `/vcpi-rp-1-1/vcpi-sub-rp`
 -  `vcpi-rp-1-1/vcpi-sub-rp/`
 - `/vcpi-rp-1-1/vcpi-sub-rp/`

I want to let user use any of the above.

src/vsphere_cpi/lib/cloud/vsphere/resources/resource_pool.rb
EleanorRigby: Yeah. My original motivation was to implement specific faults to rescue but it is not needed. Will iron this out into `if and elsif and else`

src/vsphere_cpi/lib/cloud/vsphere/resources/resource_pool.rb
EleanorRigby: Good Catch. thanks!

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
EleanorRigby: Done.

PR #202, opened 2019-03-28
WIP:Make Host Group Vm Group Affinity rule configurable for `MUST` vs `SHOULD`

EleanorRigby:
## What?
- Host group is now a Hash containing two keys:
    - Name
    - DRS Rule Type : MUST or SHOULD (Value check is not enforced)

- Create a DRS rule based on host group spec in cluster definition

- The advised values for rule type are MUST or SHOULD.

- Any bad value will make the drs rule default to `should` or `optional`

- Only a string with value MUST in any case and any number of leading or
 trailing white space characters is considered a valid value to make
 rule mandatory.

- For the must and should rule that concern with single host group, two
 different rules are created with two different vm groups tied to
 respective rules.

---
## WIP
Integration Tests

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cluster_config.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cluster_config.rb
@@ -1,6 +1,5 @@
 module VSphereCloud
   class ClusterConfig
-
     attr_reader :name
 
     def initialize(name, config_hash)
@@ -8,8 +7,21 @@ module VSphereCloud
       @config = config_hash
     end
 
-    def host_group; @config['host_group']; end
+    def host_group_name
+        @config.dig('host_group', 'name')
+    end
 
-    def resource_pool; @config['resource_pool']; end
+    # Returns default ype should if drs_rule is not specified.
+    def host_group_drs_rule
+      host_group_drs_rule = @config.dig('host_group', 'drs_rule')
+      unless host_group_drs_rule.nil?
+        return host_group_drs_rule
+      end
+      'SHOULD'
+    end
+
+    def resource_pool
+      @config['resource_pool']
+    end
   end
 end
--- a/src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
@@ -32,14 +32,14 @@ module VSphereCloud
     # @param [Vim::VirtualMachine] vm
     # @param [String] vm_group_name
     # @param [String] host_group_name
-    def add_vm_host_affinity_rule(vm_group_name, host_group_name)
+    def add_vm_host_affinity_rule(vm_group_name, host_group_name, rule_type)
       DrsLock.new(DRS_LOCK_HOST_VM_GROUP).with_drs_lock do
         rule = find_rule
         # Do not create the rule if it already exists
         unless rule
           # No error is raised if given host group does not exist,
           # it still creates the rule
-          create_vm_host_affinity_rule(vm_group_name, host_group_name)
+          create_vm_host_affinity_rule(vm_group_name, host_group_name, rule_type)
         end
       end
     end
@@ -71,11 +71,15 @@ module VSphereCloud
       add_anti_affinity_rule(VimSdk::Vim::Option::ArrayUpdateSpec::Operation::EDIT, rule_key)
     end
 
-    def create_vm_host_affinity_rule(vm_group_name, host_group_name)
+    def create_vm_host_affinity_rule(vm_group_name, host_group_name, rule_type)
       vm_host_rule_info = VimSdk::Vim::Cluster::VmHostRuleInfo.new
       vm_host_rule_info.enabled = true
-      # This is a HARD AFFINITY rule.
-      vm_host_rule_info.mandatory = true
+      # Check Hard or Soft Affinity
+      if rule_type == VSphereCloud::Resources::Cluster::CLUSTER_VM_HOST_RULE_MUST
+        vm_host_rule_info.mandatory = true
+      else
+        vm_host_rule_info.mandatory =  false
+      end
       vm_host_rule_info.name = @rule_name
       vm_host_rule_info.vm_group_name = vm_group_name
       vm_host_rule_info.affine_host_group_name = host_group_name
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
@@ -11,6 +11,8 @@ module VSphereCloud
 
       CLUSTER_VM_GROUP_SUFFIX = '_vm_group'.freeze
       CLUSTER_VM_HOST_RULE_SUFFIX = '_rule'.freeze
+      CLUSTER_VM_HOST_RULE_SHOULD = 'SHOULD'.freeze
+      CLUSTER_VM_HOST_RULE_MUST = 'MUST'.freeze
       PROPERTIES = %w(name datastore resourcePool host)
       HOST_PROPERTIES = %w(name hardware.memorySize runtime.connectionState runtime.inMaintenanceMode runtime.powerState)
       HOST_COUNTERS = %w(mem.usage.average)
@@ -75,7 +77,15 @@ module VSphereCloud
       #
       # @return [String] Name of the host Group specified under a cluster config.
       def host_group
-        config.host_group
+        config.host_group_name
+      end
+
+      # Returns a host group rule if specified in cluster config. Otherwise,
+      # it returns a nil.
+      #
+      # @return [String] Name of the host Group rule type specified under a cluster config.
+      def host_group_drs_rule
+        config.host_group_drs_rule
       end
 
       # it returns a default name created by adding suffix to host group name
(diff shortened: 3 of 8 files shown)


src/vsphere_cpi/lib/cloud/vsphere/cluster_config.rb
```diff
@@ -3,12 +3,18 @@ class ClusterConfig
 
     attr_reader :name
 
+    HOST_GROUP_ATTR       = ['host_group']
```
ktchen14: While very fancy I'm not sure that this pattern is really that helpful.

src/vsphere_cpi/lib/cloud/vsphere/cluster_config.rb
```diff
@@ -3,12 +3,18 @@ class ClusterConfig
...
+    HOST_GROUP_NAME_ATTR  = HOST_GROUP_ATTR + ['name']
+    HOST_GROUP_RULE_ATTR  = HOST_GROUP_ATTR + ['drs_rule']
+
     def initialize(name, config_hash)
       @name = name
       @config = config_hash
     end
 
-    def host_group; @config['host_group']; end
+    def host_group_name; @config.dig(*HOST_GROUP_NAME_ATTR); end
```
ktchen14: Split this onto multiple lines to avoid semicolons. There's enough logic here that a single line isn't very readable.

src/vsphere_cpi/lib/cloud/vsphere/cluster_config.rb
```diff
@@ -3,12 +3,18 @@ class ClusterConfig
...
+
+    def host_group_drs_rule; @config.dig(*HOST_GROUP_RULE_ATTR); end
```
ktchen14: Split this onto multiple lines to avoid semicolons. There's enough logic here that a single line isn't very readable.

src/vsphere_cpi/lib/cloud/vsphere/drs_rules/drs_rule.rb
```diff
@@ -71,11 +71,11 @@ def update_rule(rule_key)
...
     end
 
-    def create_vm_host_affinity_rule(vm_group_name, host_group_name)
+    def create_vm_host_affinity_rule(vm_group_name, host_group_name, rule_type)
       vm_host_rule_info = VimSdk::Vim::Cluster::VmHostRuleInfo.new
       vm_host_rule_info.enabled = true
-      # This is a HARD AFFINITY rule.
-      vm_host_rule_info.mandatory = true
+      # Check Hard or Soft Affinity
+      vm_host_rule_info.mandatory = (rule_type == VSphereCloud::Resources::Cluster::CLUSTER_VM_HOST_RULE_MUST) ? true : false
```
ktchen14: Unless `rule_type` is a very magic type then the ternary operator here is unnecessary.

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
```diff
@@ -11,6 +11,8 @@ class Cluster
 
       CLUSTER_VM_GROUP_SUFFIX = '_vm_group'.freeze
       CLUSTER_VM_HOST_RULE_SUFFIX = '_rule'.freeze
+      CLUSTER_VM_HOST_RULE_SHOULD = 'SHOULD'.freeze
```
ktchen14: Not sure either of these constants are very helpful. They're kind of just longer versions of the actual string.

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
```diff
@@ -75,17 +77,35 @@ def vm_host_affinity_rule_name
...
-        host_group + CLUSTER_VM_GROUP_SUFFIX
+        if host_group.length > 100
+          return "#{self.rule_type_suffix}-vm_group-#{(0...6).map { (97 + rand(26)).chr }.join}"
+        end
+        host_group + self.rule_type_suffix + CLUSTER_VM_GROUP_SUFFIX
+      end
+
+      def rule_type_suffix
+        rule_type = host_group_drs_rule
+        (rule_type.nil? || rule_type.strip.casecmp?(CLUSTER_VM_HOST_RULE_MUST) == false) ?
```
ktchen14: This ternary expression is too long and confusing.

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
```diff
@@ -75,17 +77,35 @@ def vm_host_affinity_rule_name
...
       end
 
       # it returns a default name created by adding suffix to host group name
       #
       # @return [String] Name of the VM Group
       def vm_group
         return nil if host_group.nil?
-        return "vm_group-#{(0...6).map { (97 + rand(26)).chr }.join}" if host_group.length > 100
-        host_group + CLUSTER_VM_GROUP_SUFFIX
+        if host_group.length > 100
```
ktchen14: I can't tell what this code is trying to do. I know the original was inscrutable as well. Can we fix it?

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
```diff
@@ -45,15 +48,19 @@
       cpi.client.cloud_searcher.get_managed_object(
         VimSdk::Vim::ClusterComputeResource, name: @second_cluster_name)
     end
-    let(:first_host_vm_group_name) { @first_host_group+VSphereCloud::Resources::Cluster::CLUSTER_VM_GROUP_SUFFIX }
-    let(:second_host_vm_group_name) { @second_host_group+VSphereCloud::Resources::Cluster::CLUSTER_VM_GROUP_SUFFIX }
-    let(:third_host_vm_group_name) { @third_host_group+VSphereCloud::Resources::Cluster::CLUSTER_VM_GROUP_SUFFIX }
+    let(:first_host_vm_group_name) { @first_host_group+'MUST'+VSphereCloud::Resources::Cluster::CLUSTER_VM_GROUP_SUFFIX }
```
ktchen14: Can we use string interpolation here instead of addition? This is kind of hard to read.

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
```diff
@@ -198,8 +226,8 @@
               datastore_pattern: @datastore_shared_pattern,
               persistent_datastore_pattern: @datastore_shared_pattern,
               clusters: [
-                { @cluster_name => { host_group: @first_host_group } },
-                { @second_cluster_name => { host_group: @third_host_group } },
+                { @cluster_name => { host_group: {'name' => @first_host_group, 'drs_rule' => 'MUST'} } },
+                { @second_cluster_name => { host_group: {'name' => @third_host_group, 'drs_rule' => 'must'} } },
```
ktchen14: Is this supposed to be lowercase?

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
```diff
@@ -160,32 +168,52 @@
...
+              vm_host_rule = get_vm_host_affinity_rule(cluster_mob, drs_rule)
+              expect(vm_host_rule.vm_group_name).to eq(first_host_vm_group_name)
+              expect(vm_host_rule.affine_host_group_name).to eq(@first_host_group)
+            end
+            expect(get_count_vm_host_affinity_rules(cluster_mob)).to eq(0)
+          end
+        end
+        context 'when the drs rule type is should' do
+          let(:drs_rule) { 'SHOULD' }
+          it 'creates VM and attach vm group host affinity rule' do
```
ktchen14: This may benefit from RSpec's shared examples as I can't tell how it differs (other than the `'SHOULD'`) from the above `it` example block:

https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
```diff
@@ -219,9 +247,9 @@
             expect(expected_host_group_mob.host.map(&:name)).to include(vm_host)
             vm_cluster_mob = vm.cluster == @cluster_name ? cluster_mob : second_cluster_mob
             expect(get_count_vm_host_affinity_rules(vm_cluster_mob)).to eq(1)
-            expect(find_vm_group_mob(expected_host_group_mob.name, vm_cluster_mob).vm).to include(vm.mob)
+            expect(find_vm_group_mob(expected_host_group_mob.name, clustermob: vm_cluster_mob).vm).to include(vm.mob)
             vm_host_rule = get_vm_host_affinity_rule(vm_cluster_mob)
-            expect(vm_host_rule.vm_group_name).to eq(expected_host_group_mob.name+VSphereCloud::Resources::Cluster::CLUSTER_VM_GROUP_SUFFIX)
+            expect(vm_host_rule.vm_group_name).to eq(expected_host_group_mob.name+default_drs_rule+VSphereCloud::Resources::Cluster::CLUSTER_VM_GROUP_SUFFIX)
```
ktchen14: I think this line is too long. We might want a helper function for our specs. And also the addition rather than string interpolation makes this difficult to understand.

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
```diff
@@ -404,11 +432,18 @@ def find_host_group_mob(host_group_name, clustermob=cluster_mob)
     end
   end
 
-  def get_vm_host_affinity_rule(cluster_mob)
-    # return ing the first as we do not expect more rules to be present while
+  def get_vm_host_affinity_rule(cluster_mob, name_hint=nil)
```
ktchen14: Spaces around `=`. No spaces is Python style.

EleanorRigby:
Closing for lack of response on `feature request`

EleanorRigby:
@sanikagawhane

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
EleanorRigby: They are used in multiple places so I wanted to keep constants which will avoid spelling mistakes by developers and thus avoid the bugs

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
EleanorRigby: The code handles these `drs_rule` strings in a case free manner. 

That's why tests mention both to ensure robustness.

src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
EleanorRigby: I shortened the constant with Another `let` but on changing it I believe `+` is neater than interpolation via `"#{something}..."`

PR #205, opened 2019-04-15
Add support for human readable names (ASCII only)

neilcx:
- Enable human readable names for BOSH created VMs
- Add Unit Tests
- Add Integration Tests

[153524141]https://www.pivotaltracker.com/n/projects/2110693/stories/153524141

# Description

Add support to enable human readable names for BOSH created VMs.
Currently only support names with legal ASCII characters.

## Type of change

- [ X] This change requires a documentation update

## How Has This Been Tested?

- [X] Unit Test
- [X] Integration Test

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -307,7 +307,12 @@ module VSphereCloud
             global_clusters: @datacenter.clusters,
             disk_configurations: disk_configs,
             storage_policy: policy_name,
+            enable_human_readable_name: config.human_readable_name_enabled?
           }
+
+          if config.human_readable_name_enabled?
+            manifest_params.update(human_readable_name_info: update_name_info_from_bosh_env(environment))
+          end
 
           vm_config = VmConfig.new(
             manifest_params: manifest_params,
@@ -871,6 +876,18 @@ module VSphereCloud
       )
       disk_configurations.push(ephemeral_disk_config)
       return disk_configurations, policy_name
+    end
+
+    # An alias to {VSphereCloud::Cloud}'s get name information from bosh environment method
+    # @param environment [Hash] the BOSH environment setting information
+    #  [Array<String>, nil] return a array with 2 strings inside if there are both instance group name and deployment name
+    #   or a nil otherwise
+    def update_name_info_from_bosh_env(environment)
+      return nil if environment.nil?
+      instance_group_name = environment.dig('bosh', 'groups', 2)
+      deployment_name = environment.dig('bosh', 'groups', 1)
+      return nil if instance_group_name.nil? || deployment_name.nil?
+      OpenStruct.new(inst_grp: instance_group_name, deployment: deployment_name)
     end
   end
 end
--- a/src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
@@ -1,3 +1,6 @@
+require 'uri'
+require 'ostruct'
+
 module VSphereCloud
   class VmConfig
     include Logger
@@ -12,7 +15,24 @@ module VSphereCloud
     end
 
     def name
-      @vm_cid ||= "vm-#{SecureRandom.uuid}"
+      return @vm_cid if defined? @vm_cid
+
+      unless human_readable_name_enabled?
+        @vm_cid = "vm-#{SecureRandom.uuid}"
+        return @vm_cid
+      end
+
+      if human_readable_name_info.nil?
+        logger.info("Not enough metadata to construct human readable name, using UUID based naming.")
+        @vm_cid = "vm-#{SecureRandom.uuid}"
+        return @vm_cid
+      end
+      @vm_cid = generate_human_readable_name(human_readable_name_info.inst_grp, human_readable_name_info.deployment)
+      if @vm_cid != "#{URI.escape(@vm_cid)}"
+        logger.info("Metadata contains non ASCII characters, using UUID based naming.")
+        @vm_cid = "vm-#{SecureRandom.uuid}"
+      end
+      @vm_cid
     end
 
     def cluster_placements
@@ -56,8 +76,33 @@ module VSphereCloud
       @manifest_params[:storage_policy]
     end
 
+    def human_readable_name_enabled?
+      @manifest_params[:enable_human_readable_name]
+    end
+
+    def human_readable_name_info
+      @manifest_params.fetch(:human_readable_name_info, nil)
+    end
+
     def networks_spec
       @manifest_params[:networks_spec] || {}
+    end
+
+    def generate_human_readable_name(instance_name, deployment_name)
+      max_prefix = 79 - 12 - 2  # vCenter size limit 80, suffix 12, underscores 2
+      uuid_suffix = SecureRandom.uuid.slice(-12, 12)
+
+      if instance_name.size + deployment_name.size > max_prefix
+        trim_codepoint_size = instance_name.size + deployment_name.size - max_prefix
+        if deployment_name.size <= 25 #  deployment name length set to 25
+          instance_name = instance_name.slice(0, instance_name.size - trim_codepoint_size)
+        else
+          instance_codepoint_size = [max_prefix -  25, instance_name.size].min
+          instance_name = instance_name.slice(0, instance_codepoint_size)
+          deployment_name = deployment_name.slice(0, max_prefix - instance_codepoint_size)
+        end
+      end
+      "#{instance_name}_#{deployment_name}_#{uuid_suffix}"
     end
 
     def vsphere_networks
--- a/src/vsphere_cpi/spec/integration/human_readable_names_spec.rb
+++ b/src/vsphere_cpi/spec/integration/human_readable_names_spec.rb
@@ -1,0 +1,122 @@
+require 'integration/spec_helper'
+
+describe 'enable human readable names' do
+  let(:vm_type) do
+    {
+        'ram' => 512,
+        'disk' => 2048,
+        'cpu' => 1,
+    }
+  end
+
+  let(:network_spec) do
+    {
+      'static' => {
+        'ip' => "169.254.#{rand(1..254)}.#{rand(4..254)}",
+        'netmask' => '255.255.254.0',
+        'cloud_properties' => {'name' => @vlan},
+        'default' => ['dns', 'gateway'],
+        'dns' => ['169.254.1.2'],
+        'gateway' => '169.254.1.3'
+      }
+    }
+  end
+
+  let(:human_readable_name_cpi) do
+    options = cpi_options({'enable_human_readable_name' => true})
+    VSphereCloud::Cloud.new(options)
+  end
+
+  let(:machine_name_cpi) do
+    options = cpi_options({'enable_human_readable_name' => false})
+    VSphereCloud::Cloud.new(options)
+  end
+
+  context 'when enable human readable names is set to false' do
+    let(:environment){ {'bosh' => { 'groups' => ['fake-director-name', 'fake-deployment-name', 'fake-instance-group-name'] } } }
+    it 'create vm with UUID based name' do
+      begin
+        test_vm_id = machine_name_cpi.create_vm(
+          'agent-007',
+          @stemcell_id,
+          vm_type,
+          network_spec,
+          [],
+          environment
+        )
+        expect(test_vm_id).to_not be_nil
+        expect(test_vm_id.size).to eq 39 # "vm_" + 36 digits UUID
+        expect(test_vm_id).to match /vm-.*/
+      ensure
+        delete_vm(machine_name_cpi, test_vm_id)
+      end
+    end
+  end
+
+  context 'when enable human readable names is set to true' do
+    context 'when instance group name and deployment name are in ASCII characters' do
+      context 'when both instance group name and deployment name are set' do
+        let(:environment){ {'bosh' => { 'groups' => ['fake-director-name', 'fake-deployment-name', 'fake-instance-group-name'] } } }
+        it 'create vm with human readable name' do
+          begin
+            test_vm_id = human_readable_name_cpi.create_vm(
+                'agent-007',
+                @stemcell_id,
+                vm_type,
+                network_spec,
+                [],
+                environment
+            )
+            expect(test_vm_id).to_not be_nil
+            expect(test_vm_id.size).to eq 58  # "fake-instance-group-name_fake-deployment-name_" + 12 digits suffix
+            expect(test_vm_id).to start_with('fake-instance-group-name_fake-deployment-name_')
+          ensure
+            delete_vm(human_readable_name_cpi, test_vm_id)
+          end
+        end
+      end
+
+      context 'when bosh environment metadata is not in correct format' do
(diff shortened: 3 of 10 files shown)


src/vsphere_cpi/lib/cloud/vsphere/file_provider.rb
```diff
@@ -43,7 +43,7 @@ def upload_file_to_url(url, body, headers)
     private
 
     def fetch_file_from_datastore_via_datacenter(datacenter_name, datastore_name, path)
-      url ="https://#{@vcenter_host}/folder/#{path}?dcPath=#{URI.escape(datacenter_name)}" +
+      url ="https://#{@vcenter_host}/folder/#{URI.escape(path)}?dcPath=#{URI.escape(datacenter_name)}" +
```
ktchen14: We have to be careful here. I'm not sure if `path` when passed into this function is already URL escaped. If it is then we'll end up double escaping this. In at least one case:

```ruby
    def get_current_env(vm, datacenter_name)
      logger.info("Getting current agent env from vm '#{vm.name}' in datacenter '#{datacenter_name}'")
      cdrom = @client.get_cdrom_device(vm)
      env_iso_folder = env_iso_folder(cdrom)
      return unless env_iso_folder

      datastore = cdrom.backing.datastore
      datastore_pattern = Regexp.escape(datastore.name)
      result = env_iso_folder.match(/\[#{datastore_pattern}\] (.*)/)
      raise Bosh::Clouds::CloudError.new("Could not find matching datastore name '#{datastore.name}'") unless result
      env_path = result[1]

      contents = @file_provider.fetch_file_from_datastore(datacenter_name, datastore, "#{env_path}/env.json")
      raise Bosh::Clouds::CloudError.new("Unable to load env.json from '#{env_path}/env.json'") unless contents

      JSON.load(contents)
    end
```

This makes me nervous that we may break existing behavior.

The same goes for the other changes of a similar nature in this file.

src/vsphere_cpi/lib/cloud/vsphere/resources/disk.rb
```diff
@@ -13,7 +15,7 @@ def initialize(size_in_mb, datastore, folder, cid)
       end
 
       def path
-        "[#{@datastore.name}] #{@folder}/#{@cid}.vmdk"
+        "[#{URI.escape(@datastore.name)}] #{URI.escape(@folder)}/#{URI.escape(@cid)}.vmdk"
```
ktchen14: This change really requires a link to some documentation or otherwise. I'm not sure if the string inside of the datastore name `[]` part of the path can be URL escaped.

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -11,8 +19,28 @@ def upgrade_hw_version?(vmtype_hw_version, global_hw_version)
       vmtype_hw_version.nil? ? global_hw_version : vmtype_hw_version
     end
 
+    # NOTE : This always creates a new name
+    def get_uuid_name
```
ktchen14: In general method names in Ruby beginning with `get_` should be avoided as they generally indicate that we're trying to do the wrong thing.

In this case we have a method called `get_uuid_name` that actually **sets** the `@vm_cid` instance variable. And in fact based on what I can tell every time this method is called `@vm_cid` is actually reset based on its return value.

We should just get rid of this method.

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -11,8 +19,28 @@ def upgrade_hw_version?(vmtype_hw_version, global_hw_version)
...
+      @vm_cid = "vm-#{SecureRandom.uuid}"
+      @vm_cid
+    end
+
+    # This returns if the name is already cached.
     def name
-      @vm_cid ||= "vm-#{SecureRandom.uuid}"
+      return @vm_cid if @vm_cid
```
ktchen14: The Ruby idiom is generally:

```ruby
def name
  return @vm_cid if defined? @vm_cid
  ...
end
```

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -11,8 +19,28 @@ def upgrade_hw_version?(vmtype_hw_version, global_hw_version)
...
+      if human_readable_name_enabled? && human_readable_name_info.has_key?('instance_group_name') && human_readable_name_info.has_key?('deployment_name')
```
ktchen14: Not sure I like this. We have to do all of these `has_key?` checks because `human_readable_name_info` is a `Hash`. We should probably just make it an actual class or a `Struct` with the `instance_group_name` and `deployment_name` required attributes.

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -11,8 +19,28 @@ def upgrade_hw_version?(vmtype_hw_version, global_hw_version)
...
+        @vm_cid = generate_human_readable_name(human_readable_name_info['instance_group_name'], human_readable_name_info['deployment_name'])
+      elsif human_readable_name_enabled?
+          logger.info("Not enough metadata to construct human readable name, using UUID based naming.")
+          @vm_cid = get_uuid_name
```
ktchen14: Putting the default case in `else` should be avoided. Generally methods should return with defaults early.

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -56,10 +84,60 @@ def storage_policy_name
       @manifest_params[:storage_policy]
     end
 
+    def human_readable_name_enabled?
+      @manifest_params[:enable_human_readable_name]
+    end
+
+    def human_readable_name_info
+      @manifest_params.fetch(:human_readable_name_info, {})
```
ktchen14: Shouldn't this default to `nil`?

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -56,10 +84,60 @@ def storage_policy_name
...
+    end
+
     def networks_spec
       @manifest_params[:networks_spec] || {}
     end
 
+    def generate_human_readable_name(instance_name, deployment_name)
```
ktchen14: This code block looks really complicated for what it's trying to do (especially as it's not going to handle Unicode). Can we simplify this?

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -1,7 +1,15 @@
+require 'URI'
+
 module VSphereCloud
   class VmConfig
     include Logger
 
+    MAX_VM_NAME = 79
```
ktchen14: We should comment these two constants to explain where we're getting them from.

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -1,7 +1,15 @@
...
+    MAX_VM_NAME_IN_BYTE = 213
+    MAX_DEPLOYMENT_NAME = 25
```
ktchen14: These three comments are basically implementation details of a private method and in my opinion don't need to be here.

src/vsphere_cpi/spec/integration/human_readable_names_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+    VSphereCloud::Cloud.new(options)
+  end
+
+  let(:machine_name_cpi) do
+    options = cpi_options({'enable_human_readable_name' => false})
+    VSphereCloud::Cloud.new(options)
+  end
+
+  context 'When enable human readable names is set to false' do
+    let(:environment){
```
ktchen14: Formatting and whitespace looks off.

src/vsphere_cpi/spec/integration/human_readable_names_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+          'agent-007',
+          @stemcell_id,
+          vm_type,
+          network_spec,
+          [],
+          environment
+        )
+        expect(test_vm_id).to_not be_nil
+        expect(test_vm_id.size).to eq 39
+        expect(test_vm_id).to start_with('vm')
```
ktchen14: We should probably use a regular expression match here instead of `start_with('vm')`. In fact we can probably replace all three lines with a single regular expression match.

We can probably even mock out the `SecureRandom.uuid` method here for an exact match.

src/vsphere_cpi/spec/integration/human_readable_names_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+          begin
+            test_vm_id = human_readable_name_cpi.create_vm(
+                'agent-007',
+                @stemcell_id,
+                vm_type,
+                network_spec,
+                [],
+                environment
+            )
+            expect(test_vm_id).to_not be_nil
```
ktchen14: Not sure these two lines here are helpful. Magic numbers especially should be avoided (what's 58)? When we use hardcoded numbers we should explain how we're getting them.

src/vsphere_cpi/spec/integration/human_readable_names_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+        end
+      end
+    end
+
+    context 'when instance group name and deployment name contain non_ASCII characters' do
+      let(:environment){
+        {'bosh' => { 'groups' => ['fake-director-name', 'ÅÅÅÅ', 'αβγδεζηθικλμνξοπρςστυφχψω'] }
+        }
+      }
+      it 'create vm with human readable name' do
```
ktchen14: Looks like the `it` examples in this file fall into only two categories:

1. It creates a VM with a human readable name
2. It creates a VM with a UUID based name

We probably want RSpec's shared examples in that case:

https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples

src/vsphere_cpi/lib/cloud/vsphere/file_provider.rb
neilcx: @ktchen14 I have tested and the `path` here is not escaped yet. Since we are not supporting non-ASCII letters now, I will roll back it ( to make minimum change to CPI). And when we trying to  support non-ASCII letters, we change it.

src/vsphere_cpi/lib/cloud/vsphere/resources/disk.rb
neilcx: @ktchen14  Here is an escaped example `"[isc-cl1-ds-0] %CE%B1%CE%B2_%C3%85%C3%85%C3%85%C3%85_4e4ae49d16fa/ephemeral_disk.vmdk"` For here, path is `αβ_ÅÅÅÅ_4e4ae49d16fa`

src/vsphere_cpi/spec/integration/human_readable_names_spec.rb
neilcx: Added comments in the code.  58 is the length of [fake-instance-group-name_fake-deployment-name_] and 12 digits random suffix

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
neilcx: Simplified this by eliminating unicode part

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -56,10 +75,35 @@ def storage_policy_name
...
       @manifest_params[:networks_spec] || {}
     end
 
+    def generate_human_readable_name(instance_name, deployment_name)
+      max_prefix = 79 - 12 - 2  # size limit 80 from vCenter, 12 digits for unique suffix, 2 for underscores
+      uuid_suffix = SecureRandom.uuid.slice(-12, 12)
+
+      if instance_name.size + deployment_name.size > max_prefix
+        trim_codepoint_size = instance_name.size + deployment_name.size - max_prefix
+        if deployment_name.size <= 25 # ideal size for deployment name set to 25
```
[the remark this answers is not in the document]
neilcx: I talked with Kaiting, he prefer to keep numbers in line and make a short comment.  I will write more detailed explanation in the document for our user.

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -56,10 +75,35 @@ def storage_policy_name
...
+    def human_readable_name_info
+      @manifest_params.fetch(:human_readable_name_info, nil)
+    end
+
     def networks_spec
       @manifest_params[:networks_spec] || {}
     end
 
+    def generate_human_readable_name(instance_name, deployment_name)
+      max_prefix = 79 - 12 - 2  # size limit 80 from vCenter, 12 digits for unique suffix, 2 for underscores
```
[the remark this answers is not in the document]
neilcx: I talked with Kaiting, he prefer to keep numbers in line and make a short comment. I will write more detailed explanation in the document for our user.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -872,5 +877,17 @@ def disk_configurations(vm_type, existing_disk_cids=[])
...
+    # An alias to {VSphereCloud::Cloud}'s get name information from bosh environment method
+    # @param environment [Hash] the BOSH environment setting information
+    #  [Array<String>, nil] return a array with 2 strings inside if there are both instance group name and deployment name
+    #   or a nil otherwise
+    def update_name_info_from_bosh_env(environment)
+      return nil if environment.nil?
+      instance_group_name = environment.dig('bosh', 'groups', 2)
+      deployment_name = environment.dig('bosh', 'groups', 1)
+      return nil if instance_group_name.nil? || deployment_name.nil?
+      [instance_group_name, deployment_name]
```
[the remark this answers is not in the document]
neilcx: Done,  updated code and tests.

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -12,7 +14,24 @@ def upgrade_hw_version?(vmtype_hw_version, global_hw_version)
...
+        @vm_cid = "vm-#{SecureRandom.uuid}"
+        return @vm_cid
+      end
+
+      if human_readable_name_info.nil?
+        logger.info("Not enough metadata to construct human readable name, using UUID based naming.")
+        @vm_cid = "vm-#{SecureRandom.uuid}"
+        return @vm_cid
+      end
+      @vm_cid = generate_human_readable_name(human_readable_name_info[0], human_readable_name_info[1])
```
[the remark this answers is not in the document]
neilcx: done.

src/vsphere_cpi/spec/integration/human_readable_names_spec.rb
neilcx: Done. I have implemented with shared examples.  Use structure like this:

when human readable name set to false
  normal way (no shared example used since only 1 test)
when set to true
  shared_example (create vm with human readable name)  
  shared_example (create vm with uuid based name)

  context 'when bosh environment metadata is not in correct format' do => uuid way
  context 'when inst_grp and deployment names contain non_ASCII characters'  => uuid way
  context 'when both names are set with ASCII characters only'  => human readable way

PR #212, opened 2019-05-21
Limited Support for NSX-T Policy API

EleanorRigby:
Limited support for NSX-T Policy API.

Major Flow
 - create segment ports
 - Tag the segment ports with `policy_api => true`
 - attach them to vnics of vm
 - clone vm
 - add vm to policy groups
 - update metadata on vm segment ports
 - Make a decision based on vm's logical ports' tags to use mgr vs policy for deletion
 - remove vm from policy groups
 - remove vm segment ports form segments
 - delete vm


src/vsphere_cpi/lib/cloud/vsphere/nsxt_policy_provider.rb
```diff
@@ -0,0 +1,360 @@
...
+      retrieve_all_with_pagination(object_type: 'Policy Group') do |cursor=nil|
+        unless cursor.nil?
+          policy_group_api.list_group_for_domain(DEFAULT_NSXT_POLICY_DOMAIN, cursor: cursor)
+        else
+          policy_group_api.list_group_for_domain(DEFAULT_NSXT_POLICY_DOMAIN)
+        end
+      end
+    end
+
+    def retry_on_conflict(log_str)
```
EleanorRigby: @nehagjain15 @ktchen14 : Need help to express this better. Want to avoid API Error due to high requests/second (`>100`). Suggestions to code it better?

PR #221, opened 2019-06-20
FIX: NSX-T client cert isn't actually sent at runtime

svrc-pivotal:
# Description

The NSX-T mTLS client cert authentication is broken at runtime as the NSXTAPIClientBuilder never properly passes the key pair in.   The unit & integration tests had manually created the ApiClient but nothing actually exercises the NSXTAPIClientBuilder which missed these properties in its constructor.

This will allow for supporting NSX_T with mTLS client cert auth.

## Type of change

## How Has This Been Tested?

I've not tested this beyond a unit test run of existing unit tests. This PR is fairly self-evident of the simple fix & missing four lines of code.  You can reject it for a more complete / tested fix if you'd like, I'm short on time but figured this would be easier than filing an issue.

# Checklist:


EleanorRigby:
Closing this as #237 solves it comprehensively.

PR #226, opened 2019-07-09
Support Shared Nothing svMotion and VMware replication disk migration 

svrc:
# Description
1. Suppresses some errors that occur when`env.iso`and `env.json` are missing at VM detach or delete time (where it doesn't really matter if they're gone at this point).   This occurs with svMotion where datastores can't be mounted across clusters (e.g. VSAN, VxFlexOS)

2. Suppresses errors that occur if an old datastore path is missing after a storage vMotion or vSphere replication / SRM.    This is common with VSAN.

3. Supports a new persistent disk migration scenario - the VMDK retains its file name but the VM it is attached to has lost its vApp property that caches the datastore disk path.  This occurs with VMware Replication / SRM, where the persistent VMDK will be moved into the VM folder but will keep its filename.   In svMotion, the VMDK is moved *and* renamed, but we retain the vApp property to where to rename it.

Persistent disks

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Unit tests only; integration test was with several live customers that used storage vmotion or vmware replication

**Test Configuration**:
* Environment Variables for Integration test: VxRail, VMware 6.7, VSAN replicated from older cluster / datastore; also VMware 7.0.x with Nutanix datastores

# Checklist:

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
@@ -9,6 +9,8 @@ module VSphereCloud
 
       NSXT_LOGICAL_SWITCH = 'nsx.LogicalSwitch'.freeze
 
+      DiskWithForcedPath = Struct.new(:disk, :path)
+
       stringify_with :cid
 
       attr_reader :mob, :cid
@@ -308,7 +310,7 @@ module VSphereCloud
         end
       end
 
-      def detach_disks(virtual_disks)
+      def detach_disks(virtual_disks, force_disk_path = nil)
         reload
         check_for_nonpersistent_disk_modes
 
@@ -322,19 +324,21 @@ module VSphereCloud
           config.device_change << Resources::VM.create_delete_device_spec(virtual_disk)
 
           disk_property = get_vapp_property_by_key(virtual_disk.key)
-          unless disk_property.nil?
-            if has_persistent_disk_property_mismatch?(virtual_disk) && !@client.disk_path_exists?(@mob, disk_property.value)
-              logger.info('Persistent disk was moved: moving disk to expected location')
-              disks_to_move << virtual_disk
-            end
+          if disk_property.nil? && !force_disk_path.nil?
+            logger.info('Persistent disk is BOSH-managed but VM has no vApp property: moving disk to current location on same datastore')
+            disks_to_move << DiskWithForcedPath.new(virtual_disk, force_disk_path)
+          elsif has_persistent_disk_property_mismatch?(virtual_disk) && !@client.disk_path_exists?(@mob, disk_property.value)
+            logger.info('Persistent disk was moved: moving disk to vApp property recorded location')
+            disks_to_move << DiskWithForcedPath.new(virtual_disk,  nil)
           end
         end
 
         @client.reconfig_vm(@mob, config)
         logger.info("Detached #{virtual_disks.size} persistent disk(s)")
 
-        move_disks_to_old_path(disks_to_move)
-
+        logger.info("Renaming #{disks_to_move.size} persistent disk(s) to proper path")
+        move_disks(disks_to_move)
+        
         logger.info('Finished detaching disk(s)')
 
         virtual_disks.each do |disk|
@@ -344,17 +348,26 @@ module VSphereCloud
         logger.debug('Finished deleting persistent disk properties from vm')
       end
 
-      def move_disks_to_old_path(disks_to_move)
-        logger.info("Renaming #{disks_to_move.size} persistent disk(s)")
-        disks_to_move.each do |disk|
-          current_datastore = disk.backing.file_name.match(/^\[([^\]]+)\]/)[1]
-          original_disk_path = get_old_disk_filepath(disk.key)
-          dest_filename = original_disk_path.match(/^\[[^\]]+\] (.*)/)[1]
-          dest_path = "[#{current_datastore}] #{dest_filename}"
+      def move_disks(disks_to_move)
+        disks_to_move.each do |disk_with_path|
+          disk = disk_with_path.disk
+          disk_path = disk_with_path.path
 
+          current_datastore = disk.backing.file_name.match(/^\[([^\]]+)\]/)[1]
+          if disk_path.nil?
+            original_disk_path = get_old_disk_filepath(disk.key)
+            dest_filename = original_disk_path.match(/^\[[^\]]+\] (.*)/)[1]
+          else
+            # Use current disk path incuded with the disk
+            dest_filename =  disk.backing.file_name.match(/^\[[^\]]+\] .*\/(.*)/)[1]
+            dest_filename = "#{disk_path}/#{dest_filename}"
+          end
+          dest_path = "[#{current_datastore}] #{dest_filename}"
+          
           @client.move_disk(datacenter_mob, disk.backing.file_name, datacenter_mob, dest_path)
         end
       end
+
 
       # Apply new storage policy to VM, system disk and ephemeral disk.
 
--- a/src/vsphere_cpi/spec/unit/cloud/vsphere/resources/vm_spec.rb
+++ b/src/vsphere_cpi/spec/unit/cloud/vsphere/resources/vm_spec.rb
@@ -442,6 +442,27 @@ describe VSphereCloud::Resources::VM, fake_logger: true do
       vm.detach_disks([disk0, disk1])
     end
 
+    context 'when a vm is missing a disk property but force_move indicates the disk to detach is bosh managed' do
+      before do
+        allow(vm_mob).to receive_message_chain(:config, :v_app_config, :property).and_return([])
+      end
+      it 'tries to move the disk to the current path' do
+        expect(client).to receive(:reconfig_vm) do |mob, spec|
+          expect(mob).to equal(vm_mob)
+          expect(spec.device_change.first.device).to eq(disk0)
+          expect(spec.device_change.first.operation).to eq(VimSdk::Vim::Vm::Device::VirtualDeviceSpec::Operation::REMOVE)
+        end
+        expect(client).to receive(:move_disk).with(
+          datacenter,
+          '[datastore x] fake-disk-path/fake-file_name.vmdk',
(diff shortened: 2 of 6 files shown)


svrc-pivotal:
I've rebased this to master and have included more tests + cases.  this helps the CPI to supports disjoint storage vmotion or vsphere replication based BOSH VM migration now (using the usual path of a bosh recreate)

svrc-pivotal:
FYI I've had 2-3 more customers in the last 2 weeks using this.  We should work on a merge plan.

msegvich:
Based on recent discussion it looks like there were no issues with merging. Any reason why that shouldn't happen?

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/vm_spec.rb
```diff
@@ -435,6 +435,27 @@
       vm.detach_disks([disk0, disk1])
     end
 
+    context 'when a vm is missing a disk property but force_move indicates the disk to detach is bosh managed' do
+      before do
+        allow(vm_mob).to receive_message_chain('config.v_app_config.property').and_return([])
```
ktchen14, line 440: Prefer separate arguments with `receive_message_chain`:

```ruby
allow(vm_mob).to receive_message_chain(:config, :v_app_config.property).and_return([])
```

src/vsphere_cpi/spec/unit/cloud/vsphere/vcenter_client_spec.rb
```diff
@@ -177,9 +194,25 @@ module VSphereCloud
         end
       end
 
+
+      context 'when file manager raises "Invalid datastore path" error for the disk' do
+        it 'does not raise error' do
+          expect(task_runner).to receive(:run) do |*args, &block|
+            expect(block.call).to eq(vmdk_task)
+          end.and_raise(RuntimeError.new('Invalid datastore path some/path'))
```
ktchen14, line 202: No need to explicitly to call `RuntimeError.new` here. Just do:

```ruby
and_raise(RuntimeError, 'Invalid datastore path some/path')
```

Or better yet:

```ruby
and_raise('Invalid datastore path some/path')
```

Since `RuntimeError` is the default exception class.

src/vsphere_cpi/spec/unit/cloud/vsphere/vcenter_client_spec.rb
```diff
@@ -177,9 +194,25 @@ module VSphereCloud
...
+
+          expect(virtual_disk_manager).to receive(:delete_virtual_disk).
+              with('[some-datastore] some/path', datacenter).
+              and_return(vmdk_task)
+          expect {
```
ktchen14, line 207: No need to `expect` ... `to_not` ... `raise_error`. Just remove the entire expectation and replace it with a comment. If an error is raised the test will fail.

ktchen14:
Just did an in depth code review. Once done I think we're good to merge. @EleanorRigby should we backport this to all supported release lines?

EleanorRigby:
I don't think there is any harm in merging this across release lines.

EleanorRigby:
Also I started a pipeline yesterday : https://ci.vcna.io/teams/vcpi/pipelines/vcpi-stu-sharedNothing to verify if the commit does not break any existing tests. It passes all 7.0 nsxt 3.0 flawlessly. 

will turn on other tests as @svrc-pivotal makes changes.

src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
```diff
@@ -297,19 +298,22 @@ def detach_disks(virtual_disks)
           config.device_change << Resources::VM.create_delete_device_spec(virtual_disk)
 
           disk_property = get_vapp_property_by_key(virtual_disk.key)
-          unless disk_property.nil?
+          if disk_property.nil? && ! force_disk_path.nil?
```
EleanorRigby, line 301: space between `!` and variable

svrc:
Working on this today!   I think our favorite customer actually wants this now.

svrc:
This came up with another customer, so I will dust off my PR and bring it up to master.

src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
```diff
@@ -331,6 +335,17 @@ def move_disks_to_old_path(disks_to_move)
         end
       end
 
+      def move_disks_to_current_path(disks_to_move, disk_path)
```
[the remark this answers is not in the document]
svrc: This one is a bit tricky, and why I split the methods in the first place.   Because if you do a blanket
```
if disk_path.nil?
  original_disk_path = get_old_disk_filepath(disk.key)
  dest_filename = original_disk_path.match(/^\[[^\]]+\] (.*)/)[1]
else
  # Use current disk path
```
This changes the logic to *always* move disks to the current disk path if it's passed in, which will happen on any CPI `detach_disk` call (the only time it's not passed in during `delete_vm` where we detach all data disks remaining, usually non-BOSH persistent disks, e.g. Kubernetes PVs).   This means the movement logic will ignore the vApp property altogether and never move disks back to their old datastore (the current behavior).

With the combined `move_disks` method, we can preserve the existing logic if we pass the `force_disk_path` (whether set or nil) in per disk to ensure the appropriate behavior.  I introduced a Struct in `vm.rb` to accomplish this.   Have a look.

julian-hj:
Hi @svrc,

We're working this issue in [tracker #179015556](https://www.pivotaltracker.com/story/show/179015556) but we have hit a stumbling block trying to reproduce the issue.

The steps we tried:
#### Case #&#x2060;1
- deploy a simple deployment with 1 VM that has no persistent disk
- In vCenter, edit the VM's device settings and modify the CD-ROM to use "client device" instead of an ISO file. (This seems to be the only way to get vCenter to allow the VM to migrate)
- in the vCenter UI, migrate the VM to a different cluster that has a different vSAN datastore, inaccessible to the first cluster. Select "data center default" as the storage policy.
- in BOSH, recreate the VM

#### Case #&#x2060;2
- deploy a simple deployment with 1 VM that has a persistent disk
- Modify the CD-ROM to use "client device"
- in the vCenter UI, migrate the VM to a different cluster that has a different vSAN datastore, inaccessible to the first cluster.
- in BOSH, recreate the VM


### Issues we ran into:
#### For case #&#x2060;1
vCenter refuses to migrate the VM unless we change the CD-ROM drive to "client device" to disassociate it from the .iso file. After making that change, and migrating the VM, BOSH recreating the VM works without errors. I am guessing from your code change that there is some other way to get vCenter to perform the vMotion without reconfiguring the CD-ROM drive first. Is that right? Or is this a different behavior in vCenter 7?

#### For case #&#x2060;2
On recreate, the vSphere CPI raises an exception trying to determine if the disk exists, apparently because the old datastore isn't accessible to the new cluster. It doesn't appear that this PR fixes the specific exception in `disk_path_exists?` we are seeing:

```
Rescued Unknown: Invalid datastore path '[WLD03-vSAN-wld03-vcenter-WLD03-cluster01-vsan01] pcf_disk'.. backtrace: /var/vcap/data/packages/vsphere_cpi/aee3cbf97aa2224a54478a55e5b213834280ff4a/lib/cloud/vsphere/retryer.rb:13:in `try'
/var/vcap/data/packages/vsphere_cpi/aee3cbf97aa2224a54478a55e5b213834280ff4a/lib/cloud/vsphere/task_runner.rb:14:in `run'
/var/vcap/data/packages/vsphere_cpi/aee3cbf97aa2224a54478a55e5b213834280ff4a/lib/cloud/vsphere/vcenter_client.rb:51:in `wait_for_task'
/var/vcap/data/packages/vsphere_cpi/aee3cbf97aa2224a54478a55e5b213834280ff4a/lib/cloud/vsphere/vcenter_client.rb:387:in `disk_path_exists?'
/var/vcap/data/packages/vsphere_cpi/aee3cbf97aa2224a54478a55e5b213834280ff4a/lib/cloud/vsphere/resources/vm.rb:301:in `block in detach_disks'
/var/vcap/data/packages/vsphere_cpi/aee3cbf97aa2224a54478a55e5b213834280ff4a/lib/cloud/vsphere/resources/vm.rb:295:in `each'
/var/vcap/data/packages/vsphere_cpi/aee3cbf97aa2224a54478a55e5b213834280ff4a/lib/cloud/vsphere/resources/vm.rb:295:in `detach_disks'
```

Is there a different configuration we should be testing? Or do we perhaps need to recover this error also, and treat it as `false`?

svrc:
Hi @julian-hj , 

1. This CD-ROM ISO problem may be a vSphere 7 issue, as it's been hard to pin down exactly the circumstances when vSphere rejects this.    I've had customer situations even last week where it refused to vMotion the VM one attempt, but on another attempt had no trouble beyond a warning in the migration wizard.   It's weird because the CD-ROM drives are disconnected, so the ISO value really should have no impact.

     - Ideally I'd like to document a script to automatically reset the CD drives for such cases, but there are some complications with this - the "govc" CLI  / govmomi SDK has a "device.cdrom.eject" but this sets it to a random/non-existent host device I've found which does not help the situation.   I think PowerShell/PowerCLI may be the only reliable way to do this (which some customers have used).   Fortunately that's now usable on Mac & Linux.   Let me see if I can attach such a script after a test in my lab.

2.  for `disk_path_exists?`  I tried to use a blanket rescue at the end to just ignore any errors and return false.  See line 422: https://github.com/cloudfoundry/bosh-vsphere-cpi-release/pull/226/files#diff-a45f9fbba19a5c41822c1b4603d399fb48f9cdfdbd445f8bb5974827bf03490b

julian-hj:
We tested this again in vSphere 6.5 and were able to migrate the VM without messing with the CD-ROM, so there's another datapoint suggesting that the refusal to migrate is a vSphere 7 prenomenon.

Patching the CPI with the changes from this PR fixed the failures and allowed `bosh recreate` to work nicely so I've merged the PR, and we will let the changes roll through our pipelines.

julian-hj:
This is shipped in v65 of the CPI.

PR #234, opened 2019-09-20
Uses volume UUIDs for ephemeral and persistent disk identification

svrc-pivotal:
# Description
Traditionally the vSphere CPI used host-relative SCSI IDs for its disk identification to the bosh agent.
This does not work in a multi-SCSI controller environment such as Kubernetes, where disk devices may
be arbitrarily changed at reboot time (e.g. /dev/sda can't be assumed to be the root disk).

The existing behavior can lead to VMs locked up and/or Kubernetes PV data loss fter a VMware HA event, power outage/reboot, etc.

This fix detects if the VM has `disk.enableUUID` enabled in its extraConfig or `vmx_options`, which ensures
stable volume UUIDs between vSphere and the guest.  It will then use that UUID in its disk
identification to the BOSH agent, which already supports UUID matching.

## Related PR and Issues
Fixes PKS-341 (private JIRA)

## Type of change

## How Has This Been Tested?

- [X] Unit tests
- [X] Manual integration tests 

**Test Configuration**:
* Hardware Requirements in ESXi: 6.5

# Checklist:

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -676,12 +676,25 @@ module VSphereCloud
       network_env
     end
 
-    def generate_disk_env(system_disk, ephemeral_disk)
-      {
+    def generate_disk_env(system_disk, ephemeral_disk, vm_config)
+
+      # When disk.enableUUID is true on the vmx options, consistent volume IDs are requested, and we can use them
+      # to ensure the precise ephemeral volume is mounted.   This is mandatory for
+      # cases where multiple SCSI controllers are present on the VM, as is common with Kubernetes VMs.
+
+      if vm_config.vmx_options['disk.enableUUID'] == "1"
+        {
+          'system' => system_disk.unit_number.to_s,
+          'ephemeral' => { 'id' => ephemeral_disk.backing.uuid.downcase }, 
+          'persistent' => {}
+        }
+      else
+        {
         'system' => system_disk.unit_number.to_s,
         'ephemeral' => ephemeral_disk.unit_number.to_s,
         'persistent' => {}
-      }
+        }
+      end
     end
 
     def generate_agent_env(name, vm, agent_id, networking_env, disk_env)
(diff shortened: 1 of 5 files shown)


src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -835,7 +843,21 @@ def get_vm_env_datastore(vm)
...
       env = @agent_env.get_current_env(vm.mob, @datacenter.name)
+
       env['disks']['persistent'][director_disk_cid.raw] = device_unit_number.to_s
+
+      for options in vm.extra_config
+        logger.info("adding disk to env, looking for diskUUID on VM: #{options.key}, #{options.value}")
+        if options.key == 'disk.enableUUID'
+          logger.info("adding disk to env, FOUND disk.enableUUID")
+          if options.value == "TRUE"
+            uuid = vm.persistent_disks[0].backing.uuid.downcase
```
ktchen14: Are we sure that this is the correct persistent disk? Seems like this will associate the wrong disk when multiple persistent disks are attached to the VM.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -676,12 +676,20 @@ def generate_network_env(devices, networks, dvs_index)
       network_env
     end
 
-    def generate_disk_env(system_disk, ephemeral_disk)
-      {
+    def generate_disk_env(system_disk, ephemeral_disk, vm_config)
+      if vm_config.vmx_options['disk.enableUUID'] == "1"
```
ktchen14: Not sure that I love the idea of looking inside of the `vmx_options` hash.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
svrc-pivotal: I need to change this to go through the entire array of persistent disks as (though this is not commercially used) BOSH supports multiple persistent disks.   But we can be sure that they're BOSH persistent disks and not Kubernetes because the `vm.persistent_disks` method only returns devices that match the vApp property that demarcates BOSH persistent disks.

EleanorRigby: `vm.persistent_disks`  does *NOT only* return devices that match vApp. It returns more than that.
```
device.backing.disk_mode == Vim::Vm::Device::VirtualDiskOption::DiskMode::INDEPENDENT_PERSISTENT ||
            persistent_device_keys.any? { |key| device.key == key }
```
Check the OR condition above ^^

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
svrc-pivotal: vSphere does not normally allocate stable UUIDs on the backing disk of a volume unless enableUUID is turned on.   This approach (and/or going directly into the VIM API extraConfig as I do in the case of persistent disks) is the only consistent way of noting that this attribute is set on the VM -- all Kubo/CFCR BOSH worker VMs have a VM extension to set this as Kubernetes vSphere Cloud Provder requires it so it can manage its PVs properly.    

We could alternatively introduce a new first-class cloud property on the VM type that would reflect the underlying VMX option, as an alternative.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
svrc-pivotal: Good catch.   So I think the better approach is to introduce a new method accessor (`vm.bosh_persistent_disks` )? that filters to ONLY the BOSH persistent disks.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
ktchen14: @svrc-pivotal Yeah that summary would be very helpful. I don't want anyone in the future to remove this code so the context here is very important.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
ktchen14: Hmm @EleanorRigby is the `add_disk_to_agent_env` method even useful? Seems like it's only called from `attach_disk` in `cloud.rb` and I'm not sure that it represents a meaningful abstraction. If we inline it we'll have access to the underlying `Device` itself from which we can probably get the UUID more reliably.

svrc-pivotal:
I've added unit tests to the commit, and have cleaned the code to support multiple persistent disks - the **add_agent_to_env** method matches the disk CID being attached to the disk CID in the device list.

EronWright:
Successor: #240

PR #238, opened 2019-11-05
Transfer files only through legal hosts

EleanorRigby:
# Description

Right now CPI transfers files to/from Datastores through an ESXi host system. If this fails, it falls back to vCenter.

While transferring, CPI must pick a host that is in the ambit of BOSH orchestration. Without this change, CPI picks any random host that datastore can access. This host can be outside of designated CPI clusters. Due to security concerns , the fix enables CPI to pick only those hosts which are relevant to CPI's vSphere clusters. 

## Type of change
Bug Fix

## How Has This Been Tested?
WIP


src/vsphere_cpi/lib/cloud/vsphere/agent_env.rb
```diff
@@ -21,8 +21,8 @@ def get_current_env(vm, datacenter_name)
       result = env_iso_folder.match(/\[#{datastore_pattern}\] (.*)/)
       raise Bosh::Clouds::CloudError.new("Could not find matching datastore name '#{datastore.name}'") unless result
       env_path = result[1]
-
-      contents = @file_provider.fetch_file_from_datastore(datacenter_name, datastore, "#{env_path}/env.json")
+      ds_hosts = get_vm_vsphere_cluster_hosts(vm)
```
ktchen14: To get the list of datastore accessible hosts should we just use the `Datastore.host` property on the datastore itself? Then we can avoid the extra parameter to `fetch_file_from_datastore` as well.

EleanorRigby: The name is bit misleading. These are the hosts that are accessible from DS but only those hosts which are part of VM's parent cluster.

ktchen14: Maybe I need some more context then. Why is this important? Shouldn't we be able to upload or retrieve through any host that can access the datastore?

EleanorRigby: I send in a list of hosts to file provider and file provider module selects one healthy host out of them.

EleanorRigby: This is the context :

While transferring, CPI must pick a host that is in the ambit of BOSH orchestration. Without this change, CPI picks any random host that datastore can access. This host can be outside of designated CPI clusters. Due to security concerns , the fix enables CPI to pick only those hosts which are relevant to CPI's vSphere clusters.

EleanorRigby: There might `not` exist a way we can do that. There are two sources for cluster info : Global and Cloud Config.

We will be able to tell about global CPI clusters from the properties but we might not have the cloud-config in case of `attach_disk` which also requires file upload.

The global config defined clusters might not have access to a DS where we are uploading. This will happen if a totally new cluster is added as part of the cloud-config with a DS to which it solely has connectivity. Then we will fail to find a host which can transfer files.


The only source of truth I can bank on in this flow is the VM for which files are being uploaded. Traveling up the chain of parents for that VM will give me a cluster guaranteed to be in BOSH's orchestration ambit.

PR #240, opened 2019-11-07
Uses volume UUIDs for ephemeral and persistent disk identification

EleanorRigby:
# Description
Traditionally the vSphere CPI used host-relative SCSI IDs for its disk identification to the bosh agent.
This does not work in a multi-SCSI controller environment such as Kubernetes, where disk devices may
be arbitrarily changed at reboot time (e.g. /dev/sda can't be assumed to be the root disk).

The existing behavior can lead to VMs locked up and/or Kubernetes PV data loss fter a VMware HA event, power outage/reboot, etc.

This fix detects if the VM has `disk.enableUUID` enabled in its extraConfig or `vmx_options`, which ensures
stable volume UUIDs between vSphere and the guest.  It will then use that UUID in its disk
identification to the BOSH agent, which already supports UUID matching.

## Related PR and Issues
Fixes PKS-341 (private JIRA)

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [X] Unit tests
- [X] Manual integration tests 

**Test Configuration**:
* Hardware Requirements in ESXi: 6.5

# Checklist:

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -697,12 +697,26 @@ module VSphereCloud
       network_env
     end
 
-    def generate_disk_env(system_disk, ephemeral_disk)
-      {
-        'system' => system_disk.unit_number.to_s,
-        'ephemeral' => ephemeral_disk.unit_number.to_s,
-        'persistent' => {}
-      }
+    def generate_disk_env(system_disk, ephemeral_disk, vm_config)
+
+      # When disk.enableUUID is true on the vmx options, consistent volume IDs are requested, and we can use them
+      # to ensure the precise ephemeral volume is mounted.   This is mandatory for
+      # cases where multiple SCSI controllers are present on the VM, as is common with Kubernetes VMs.
+      if vm_config.vmx_options['disk.enableUUID'] == "1"
+        logger.info("Using ephemeral disk UUID #{ephemeral_disk.backing.uuid.downcase}")
+        {
+          'system' => system_disk.unit_number.to_s,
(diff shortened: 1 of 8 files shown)


EleanorRigby:
@svrc-pivotal ^^ : I have refactored it a bit and adding more tests. Closing the old one.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -697,23 +697,22 @@ def generate_network_env(devices, networks, dvs_index)
...
+    def generate_disk_env(system_disk, ephemeral_disk, disk_uuid_enabled)
 
       # When disk.enableUUID is true on the vmx options, consistent volume IDs are requested, and we can use them
       # to ensure the precise ephemeral volume is mounted.   This is mandatory for
       # cases where multiple SCSI controllers are present on the VM, as is common with Kubernetes VMs.
-
-      if vm_config.vmx_options['disk.enableUUID'] == "1"
+      if disk_uuid_enabled
         {
           'system' => system_disk.unit_number.to_s,
```
EleanorRigby: @svrc-pivotal : I am trying to understand the reason behind retaining the unit number for the system disk. Is it irrelevant to the issue or is it required?

svrc-pivotal: Good question @eleanorigby .. the bosh agent requires it to be a string value, either the /dev/sdX device or unit number.  It will fail parsing if it’s a JSON string containing the uuid.

EleanorRigby: @svrc-pivotal : Got it. It should continue to work then

PR #242, opened 2019-11-15
WIP: Remove only relevant VM from server pool

EleanorRigby:
# Description

- The delete_vm call only gets vm_cid from BOSH and not the cloud properties
        which contains details of all server pools to which the VM should be added too.
        DELETE_VM call in the absence of any cloud properties,
        does not know anything about pool members other than their IP.
        This IP is got from VC by searching the name of vm.
        So it goes on to remove that IP from all server pools which is
      - Only adding IP might result in deleting a wrong VM with same IP
        from the server pool. Adding a display_name == vm.cid provides for
        more context when deleting.
        wrong.

NSXT Provider 

## Type of change

## How Has This Been Tested?
- [ ] New Unit Test in nsxt_provider_spec.rb
- [ ] New Integration Test nsxt_provider_spec.rb

# Checklist:

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -455,7 +455,6 @@ module VSphereCloud
         logger.info("Deleting vm: #{vm_cid}")
 
         vm = vm_provider.find(vm_cid)
-        vm_ip = vm.mob.guest&.ip_address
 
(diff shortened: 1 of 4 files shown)


EleanorRigby:
Will break existing deployments unless we can figure out a way to find a VM IP in the server pool and correctly identify it to a particular VM on VC.

The current method relies on display_name which might not be there in old deployments. Also, display_name seems like a property that can be edited from UI and seems not fit to be used for permanent marking.

Issue #224, opened 2019-07-02
NSX 6.4.5 -> Failed to create Security Group with unknown NSX Error: '{"errorCode":210,"details":"Another object with same name : p-bosh already exists in the current scope : globalroot-0.","rootCauseString":null,"moduleName":"core-services","errorData":null}'


agregory999:
**Describe the bug**
After upgrading NSX-V to 6.4.5 we cannot add VMs to the foundation.  We get the 210 error raised and cannot proceed.

**To Reproduce**
1. Upgrade NSX-V to 6.4.5
2. Modify PAS to add VMs to any component
3. Apply change and look at CPI logs

**CPI Error Log**
/var/vcap/data/packages/vsphere_cpi/ae7eb89db6e1d4d47b4c68b759eae13f31620669/lib/ruby_vim_sdk/vmomi_support.rb:56: warning: constant ::Fixnum is deprecated
I, [2019-07-02T19:30:05.297445 #4562]  INFO -- [req_id cpi-449847]: Starting info...
D, [2019-07-02T19:30:05.298991 #4562] DEBUG -- [req_id cpi-449847]: Running method 'RetrieveServiceContent'...
D, [2019-07-02T19:30:05.387624 #4562] DEBUG -- [req_id cpi-449847]: Running method 'Login'...
I, [2019-07-02T19:30:05.433187 #4562]  INFO -- [req_id cpi-449847]: Finished info in 0.14 seconds
D, [2019-07-02T19:30:05.433610 #4562] DEBUG -- [req_id cpi-449847]: Running method 'Logout'...
/var/vcap/data/packages/vsphere_cpi/ae7eb89db6e1d4d47b4c68b759eae13f31620669/lib/ruby_vim_sdk/vmomi_support.rb:56: warning: constant ::Fixnum is deprecated
I, [2019-07-02T19:30:05.337629 #4556]  INFO -- [req_id cpi-427359]: Starting info...
D, [2019-07-02T19:30:05.345628 #4556] DEBUG -- [req_id cpi-427359]: Running method 'RetrieveServiceContent'...
D, [2019-07-02T19:30:05.431371 #4556] DEBUG -- [req_id cpi-427359]: Running method 'Login'...
I, [2019-07-02T19:30:05.478331 #4556]  INFO -- [req_id cpi-427359]: Finished info in 0.14 seconds
[...]

PR #243, opened 2019-11-20
Fix for using NSX 6.4.5+. See issues/224

thelangley:
# Description

We found that a specific API endpoint was not responding in a way that we had come to expect.

This resulted in the issue 224 which was reported a while back.

Upon further investigation VMware release best practice on how to communicate with the API which basically involves specifying the Content-Type and Accept whenever you interact. This ensures that you either get xml or json responses as required.

## Related PR and Issues
Fixes https://github.com/cloudfoundry/bosh-vsphere-cpi-release/issues/224

None

## Type of change

## How Has This Been Tested?

We have used these tweaks to generate a bosh deployment which we can manage when using NSX 6.4.6.


EleanorRigby:
#244

PR #249, opened 2020-02-04
Add a device key generator to provide device keys on VM creation and Disk attachment

EleanorRigby:
# Description

The newer version of VC mandates passing unique temporary keys for device change in vSphere VM. This change adds a keygen (seeded with default device keys) to provide available unique key num.

## Type of change

## How Has This Been Tested?
 - Unit Tests


src/vsphere_cpi/lib/cloud/vsphere.rb
```diff
@@ -69,6 +69,8 @@
 require 'cloud/vsphere/disk_placement_selection_pipeline'
 require 'cloud/vsphere/vm_placement_selection_pipeline'
 require 'cloud/vsphere/pbm'
+require 'cloud/vsphere/device_key_gen'
```
ktchen14: Just spell it out `cloud/vsphere/device_key_generator`.

src/vsphere_cpi/lib/cloud/vsphere/device_key_gen.rb
```diff
@@ -0,0 +1,15 @@
+module VSphereCloud
+  module DeviceKeyGenerator
+    def self.init(keys: nil)
```
ktchen14: It looks like you want a singleton rather than a module here (if it's going to have an `init` method). Something like:

```ruby
require 'singleton'

module VSphereCloud
  class DeviceKeyGenerator
    def initialize(keys: nil)
      ...
    end
  end
end
```

Also are you sure you want the keyword argument `keys`? Seems like a regular argument is just as good here?

src/vsphere_cpi/lib/cloud/vsphere/device_key_gen.rb
```diff
@@ -0,0 +1,15 @@
...
+      @memo = Set.new(keys)
+      @seed = -1
+    end
+
+    def self.get_device_key
```
ktchen14: Should probably be just `device_key`

src/vsphere_cpi/lib/cloud/vsphere/device_key_gen.rb
```diff
@@ -0,0 +1,15 @@
...
+      until @memo.add?(@seed) != nil
```
ktchen14: Probably should be `while @memo.add?(@seed).nil?` to use `#nil?` rather than `== nil`

src/vsphere_cpi/lib/cloud/vsphere/device_key_gen.rb
```diff
@@ -0,0 +1,15 @@
...
+        @seed -= 1
+      end
+      return @seed
```
ktchen14: In fact the entire method can be this single line:

```ruby
@seed -= 1 while @memo.add?(@seed).nil?
```

src/vsphere_cpi/spec/unit/cloud/vsphere/device_key_gen_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'spec_helper'
+
+module VSphereCloud
+  describe VSphereCloud::DeviceKeyGenerator, fake_logger: true do
+    subject { described_class }
```
ktchen14: Yeah this really leads me to believe that this should be a singleton class.

src/vsphere_cpi/spec/unit/cloud/vsphere/device_key_gen_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    describe '#init' do
+      context 'when no device keys are passed' do
+        it 'initializes the memo with empty set passed and seeds it with -1' do
+          subject.init(keys: nil)
+          expect(subject.instance_variable_get("@memo")).to eq(Set.new())
```
ktchen14: Use `:@memo` rather than `"@memo"`. Also I think that this should be:

```ruby
expect(subject.instance_variable_get(:@memo)).to be_empty
```

src/vsphere_cpi/spec/unit/cloud/vsphere/device_key_gen_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+          expect(subject.instance_variable_get("@seed")).to eq(-1)
+        end
+      end
+      it 'initializes the memo with keys passed and seeds it with -1' do
+        subject.init(keys: [1,2,3])
+        expect(subject.instance_variable_get("@memo")).to eq(Set.new([1,2,3]))
+        expect(subject.instance_variable_get("@seed")).to eq(-1)
+
```
ktchen14: Extra newline.

EleanorRigby:
Closing this. Reimplement as a method in VM resource.

PR #254, opened 2020-02-12
Re write device keys for device change when cloning or reconfiguring VM

EleanorRigby:
- To support VC 7.0
     - VC 7.0 requires a unique temp key for each device we add to a VM.
     - Older CPI release used to add all devices with key as -1
     - Change to generate new temp key(negative integer) for each new
       device. Will account for already added device & their keys to a stemcell/VM.
     - Change accounts for create_vm (New NIC & Ephemeral Disk)
     - Change accounts for attach_disk (New Persistent Disk)

[#171032557](https://www.pivotaltracker.com/story/show/171032557)

### DIFF
--- a/ci/config.rb
+++ b/ci/config.rb
@@ -49,3 +49,12 @@ $pipeline.pool('6.7-NSXT24') do |pool|
     NSXT_SKIP_SSL_VERIFY: "true"
   }
 end
+
+$pipeline.pool('7.0-NSXT30') do |pool|
+  pool.params = {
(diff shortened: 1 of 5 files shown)


ktchen14:
I thought we were hanging this to be a step that rewrites the device keys after the device change list is generated. Or did that strategy not work out?

EleanorRigby:
I did not document it then and I changed my approach. It will totally work.
Let me recommit it quick. I haven't written any test yet so should be fast
to change.

PR #260, opened 2020-03-27
WIP : Add retries for adding/removing VMs from nsgroups.

SanikaGawhane:
# Description
Added retries for add and remove VMs from nsgroups.
Added corresponding unit tests.

## Related PR and Issues
Fixes [https://pivotal-spike.atlassian.net/browse/PKS-607](url)

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Toolchain:
* SDK:

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
@@ -85,8 +85,8 @@ module VSphereCloud
     end
   end
 
-  class NSXTApiCallError < StandardError
-    def initialize(error)
+  class NSXTSecGrpUpdateError < StandardError
+    def initialize(*error)
       @error = error
     end
   end
@@ -124,15 +124,13 @@ module VSphereCloud
     def add_vm_to_nsgroups(vm, ns_groups)
       return if ns_groups.nil? || ns_groups.empty?
       return if nsxt_nics(vm).empty?
-
+      lports = logical_ports(vm)
       logger.info("Adding vm '#{vm.cid}' to NSGroups: #{ns_groups}")
       Bosh::Retryable.new(
           tries: 5,
-          sleep: ->(try_count, retry_exception) { 2 },
-          on: [NSXTApiCallError]
+          on: [NSXTSecGrpUpdateError]
       ).retryer do |i|
         nsgroups = retrieve_nsgroups(ns_groups)
-        lports = logical_ports(vm)
         nsgroups.each do |nsgroup|
           logger.info("Adding LogicalPorts: #{lports.map(&:id)} to NSGroup '#{nsgroup.id}'")
           begin
@@ -142,7 +140,7 @@ module VSphereCloud
               'ADD_MEMBERS'
             )
           rescue NSXT::ApiCallError => e
-            retry if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+            raise NSXTSecGrpUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
             raise e
           end
         end
@@ -151,13 +149,12 @@ module VSphereCloud
 
     def remove_vm_from_nsgroups(vm)
       return if nsxt_nics(vm).empty?
+      lports = logical_ports(vm)
+      lport_ids = lports.map(&:id)
       Bosh::Retryable.new(
           tries: 5,
-          sleep: ->(try_count, retry_exception) { 2 },
-          on: [NSXTApiCallError]
+          on: [NSXTSecGrpUpdateError]
       ).retryer do |i|
-        lports = logical_ports(vm)
-        lport_ids = lports.map(&:id)
         nsgroups = retrieve_all_ns_groups_with_pagination.select do |nsgroup|
           nsgroup.members&.any? do |member|
             member.is_a?(NSXT::NSGroupSimpleExpression) &&
@@ -175,7 +172,7 @@ module VSphereCloud
               'REMOVE_MEMBERS'
             )
           rescue NSXT::ApiCallError => e
-            retry if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+            raise NSXTSecGrpUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
             raise e
           end
         end
--- a/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
@@ -154,11 +154,16 @@ module VSphereCloud
 
       nsgroups.each do |nsgroup|
         logger.info("Removing LogicalPorts: #{lport_ids} to NSGroup '#{nsgroup.id}'")
-        grouping_obj_svc.add_or_remove_ns_group_expression(
-          nsgroup.id,
-          *to_simple_expressions(lports),
-          'REMOVE_MEMBERS'
-        )
+        begin
+          grouping_obj_svc.add_or_remove_ns_group_expression(
+            nsgroup.id,
+            *to_simple_expressions(lports),
+            'REMOVE_MEMBERS'
+          )
+        rescue NSXT::ApiCallError => e
+          retry if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+          raise e
+        end
       end
     end
 
--- a/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
@@ -125,11 +125,16 @@ module VSphereCloud
       lports = logical_ports(vm)
       nsgroups.each do |nsgroup|
         logger.info("Adding LogicalPorts: #{lports.map(&:id)} to NSGroup '#{nsgroup.id}'")
-        grouping_obj_svc.add_or_remove_ns_group_expression(
-          nsgroup.id,
-          *to_simple_expressions(lports),
-          'ADD_MEMBERS'
-        )
+        begin
+          grouping_obj_svc.add_or_remove_ns_group_expression(
+            nsgroup.id,
+            *to_simple_expressions(lports),
+            'ADD_MEMBERS'
+          )
+        rescue NSXT::ApiCallError => e
+          retry if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+          raise e
+        end
       end
     end
 
--- a/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
@@ -86,8 +86,6 @@ module VSphereCloud
   end
 
   class NSXTOptimisticUpdateError < StandardError
-    def initialize()
-    end
   end
 
   class NSXTProvider
@@ -132,16 +130,14 @@ module VSphereCloud
             tries: 50,
             on: [NSXTOptimisticUpdateError]
         ).retryer do |i|
-          begin
-            grouping_obj_svc.add_or_remove_ns_group_expression(
-              nsgroup.id,
-              *to_simple_expressions(lports),
-              'ADD_MEMBERS'
-            )
-          rescue NSXT::ApiCallError => e
-            raise NSXTOptimisticUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
-            raise e
-          end
+          grouping_obj_svc.add_or_remove_ns_group_expression(
+            nsgroup.id,
+            *to_simple_expressions(lports),
+            'ADD_MEMBERS'
+          )
+        rescue NSXT::ApiCallError => e
+          raise NSXTOptimisticUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+          raise
         end
       end
     end
@@ -163,16 +159,14 @@ module VSphereCloud
             tries: 50,
             on: [NSXTOptimisticUpdateError]
         ).retryer do |i|
-          begin
-            grouping_obj_svc.add_or_remove_ns_group_expression(
-              nsgroup.id,
-              *to_simple_expressions(lports),
-              'REMOVE_MEMBERS'
-            )
-          rescue NSXT::ApiCallError => e
-            raise NSXTOptimisticUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
-            raise e
-          end
+          grouping_obj_svc.add_or_remove_ns_group_expression(
+            nsgroup.id,
+            *to_simple_expressions(lports),
+            'REMOVE_MEMBERS'
+          )
+        rescue NSXT::ApiCallError => e
+          raise NSXTOptimisticUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+          raise
         end
       end
     end
--- a/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
@@ -85,6 +85,12 @@ module VSphereCloud
     end
   end
 
+  class NSXTApiCallError < StandardError
+    def initialize(error)
+      @error = error
+    end
+  end
+
   class NSXTProvider
     include Logger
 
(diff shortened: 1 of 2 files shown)


src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -125,11 +125,16 @@ def add_vm_to_nsgroups(vm, ns_groups)
...
+        begin
+          grouping_obj_svc.add_or_remove_ns_group_expression(
+            nsgroup.id,
+            *to_simple_expressions(lports),
+            'ADD_MEMBERS'
+          )
+        rescue NSXT::ApiCallError => e
+          retry if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+          raise e
+        end
```
EleanorRigby: My only concern here is that there is no wait time between retries. This can severely overwhelm NSXT manager if lot of these calls are in play.
I know we decided initially to use simple retry, but can we implement this using bosh retryable and wait times between retries. like we use in function logical_ports?

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -149,11 +154,16 @@ def remove_vm_from_nsgroups(vm)
...
+        begin
+          grouping_obj_svc.add_or_remove_ns_group_expression(
+            nsgroup.id,
+            *to_simple_expressions(lports),
+            'REMOVE_MEMBERS'
+          )
+        rescue NSXT::ApiCallError => e
+          retry if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+          raise e
+        end
```
EleanorRigby: Same as above.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
SanikaGawhane: Would adding a `sleep(certain duration)` after rescuing the error and before retrying work?
Does using BOSH retryable give us some advantage over this?

EleanorRigby:
BOSH retyable is a complete code framework and will help you a write a
maintainable solution for future. If you need to add retries on more
failures, it will be easier and cleaner to do with it.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -85,6 +85,12 @@ def to_s
     end
   end
 
+  class NSXTApiCallError < StandardError
```
EleanorRigby, line 88: let's name this as NSXTSecGrpUpdateError

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -120,40 +126,59 @@ def add_vm_to_nsgroups(vm, ns_groups)
...
+        nsgroups.each do |nsgroup|
+          logger.info("Adding LogicalPorts: #{lports.map(&:id)} to NSGroup '#{nsgroup.id}'")
+          begin
+            grouping_obj_svc.add_or_remove_ns_group_expression(
+              nsgroup.id,
+              *to_simple_expressions(lports),
+              'ADD_MEMBERS'
+            )
+          rescue NSXT::ApiCallError => e
+            retry if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
```
EleanorRigby, line 145: The retryer block won't kick in unless we raise this error what it retries on.
`raise NSXTSecGrpUpdateError if e.code == 409 || e.code == 412`

EleanorRigby: We do not need to use ruby's inbuilt `retry`

Can you look into how to make this exponentially backed off? `sleep: ->(try_count, retry_exception) { 2 },`

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -120,40 +126,59 @@ def add_vm_to_nsgroups(vm, ns_groups)
...
+          logger.info("Removing LogicalPorts: #{lport_ids} to NSGroup '#{nsgroup.id}'")
+          begin
+            grouping_obj_svc.add_or_remove_ns_group_expression(
+              nsgroup.id,
+              *to_simple_expressions(lports),
+              'REMOVE_MEMBERS'
+            )
+          rescue NSXT::ApiCallError => e
+            retry if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+            raise e
```
EleanorRigby, line 179: same as above

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
SanikaGawhane: `retryable.rb` has `sleep` parameter set to default as an exponential backoff. So if we do not pass it explicitly in the `Bosh::Retryable.new()`, it will execute the default option function `expontential sleeper`

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -85,6 +85,12 @@ def to_s
     end
   end
 
+  class NSXTSecGrpUpdateError < StandardError
+    def initialize(*error)
```
EleanorRigby, line 89: Why args in the constructor? not needed, I guess

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -118,42 +124,58 @@ def initialize(client, default_vif_type)
...
+        nsgroups.each do |nsgroup|
+          logger.info("Adding LogicalPorts: #{lports.map(&:id)} to NSGroup '#{nsgroup.id}'")
+          begin
+            grouping_obj_svc.add_or_remove_ns_group_expression(
+              nsgroup.id,
+              *to_simple_expressions(lports),
+              'ADD_MEMBERS'
+            )
+          rescue NSXT::ApiCallError => e
+            raise NSXTSecGrpUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
```
EleanorRigby, line 143: not sure if this will work. Shouldn't it be `raise NSXTSecGrpUpdateError.new("NSXT 409 db conflict or 412 revision conflict encountered with error :  #{e}")`

EleanorRigby: Your code is raising a class without instantiating an object with `.new`. May be this works but I haven't seen before. Can you confirm?

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -118,42 +124,58 @@ def initialize(client, default_vif_type)
...
+        nsgroups.each do |nsgroup|
+          logger.info("Removing LogicalPorts: #{lport_ids} to NSGroup '#{nsgroup.id}'")
+          begin
+            grouping_obj_svc.add_or_remove_ns_group_expression(
+              nsgroup.id,
+              *to_simple_expressions(lports),
+              'REMOVE_MEMBERS'
+            )
+          rescue NSXT::ApiCallError => e
+            raise NSXTSecGrpUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
```
EleanorRigby, line 175: Same as above

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -118,42 +124,58 @@ def initialize(client, default_vif_type)
...
-      nsgroups.each do |nsgroup|
-        logger.info("Adding LogicalPorts: #{lports.map(&:id)} to NSGroup '#{nsgroup.id}'")
-        grouping_obj_svc.add_or_remove_ns_group_expression(
-          nsgroup.id,
-          *to_simple_expressions(lports),
-          'ADD_MEMBERS'
-        )
+      logger.info("Adding vm '#{vm.cid}' to NSGroups: #{ns_groups}")
+      Bosh::Retryable.new(
+          tries: 5,
```
EleanorRigby, line 130: Why just 5 retries?

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -118,42 +124,58 @@ def initialize(client, default_vif_type)
...
+          on: [NSXTSecGrpUpdateError]
```
EleanorRigby, line 131: What's the default sleep behavior since you did not specify in the code?

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
SanikaGawhane: I looked into it and came across this - 
[https://stackoverflow.com/questions/9638475/raising-an-exception-use-instance-or-class](url)
I think this works fine. Do you think it would be cleaner to use an instance instead?

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
SanikaGawhane: `BOSH::Retryable` requires to specify the number of tries (default number of tries = 2)
I am trying to think of a way to model this to have infinite retries.
If not, can we use a high number - say 50 or 100? Would that be an enough number of retries?

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
SanikaGawhane: `retryable.rb` has sleep parameter set to default as an exponential backoff. So if we do not pass it explicitly in the `Bosh::Retryable.new()`, it will execute the default option function expontential sleeper.

```
def default_options
      {
        tries: 2,
        sleep: exponential_sleeper,
        on: [],
        matching: /.*/,
        ensure: Proc.new {},
      }
    end
```

```
def exponential_sleeper
  lambda { |tries, _| [2**(tries-1), 10].min } # 1, 2, 4, 8, 10, 10..10 seconds
end
```

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -85,6 +85,12 @@ def to_s
     end
   end
 
+  class NSXTSecGrpUpdateError < StandardError
```
ktchen14, line 88: Ruby generally prefers whole words. This should be `NSXTSecurityGroupUpdateError`. On the other hand if this error is being used for 409/412, maybe we should make it generic? `NSXTOptimisticUpdateError`?

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
ktchen14: In general in Ruby when raising an error `.new` shouldn't be used:

```ruby
raise NSXTSecGrpUpdateError, "..."
```

Raises an instance of `NSXTSecGrpUpdateError` with `message` set to "...". Here's the documentation for the `raise` method:

https://ruby-doc.org/core-2.7.0/Kernel.html#method-i-raise

Specifically the 2+ argument form of `raise` instantiates an instance of the error class if necessary and then sets the `mesg` instance variable on that instance to the second argument. This is what allows `Exception#to_s` to get the exception message:

https://github.com/ruby/ruby/blob/9e6e39c3512f7a962c44dc3729c98a0f8be90341/error.c#L1054

The crazy thing is that the `mesg` instance variable *is not* `@mesg`. This is one of those times in which the Ruby interpreter uses an instance variable name that *doesn't* begin with `@`. So there's actually no way for you to access it from Ruby land (you can get to it in C land).

Anyways this is a very long tangent to say that the preferred way in Ruby to raise an exception with a specific error message is with the 2+ argument form of `raise`.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -118,42 +124,58 @@ def initialize(client, default_vif_type)
...
+      logger.info("Adding vm '#{vm.cid}' to NSGroups: #{ns_groups}")
+      Bosh::Retryable.new(
+          tries: 5,
+          on: [NSXTSecGrpUpdateError]
+      ).retryer do |i|
+        nsgroups = retrieve_nsgroups(ns_groups)
+        nsgroups.each do |nsgroup|
+          logger.info("Adding LogicalPorts: #{lports.map(&:id)} to NSGroup '#{nsgroup.id}'")
+          begin
+            grouping_obj_svc.add_or_remove_ns_group_expression(
```
ktchen14, line 137: I think there's a big problem here. If we hit HTTP 409/412 this code will *retry* all of the NSGroup modifications instead of the single one that failed. I'm not sure that `ADD_MEMBERS` is idempotent so we may end up adding the wrong members to the NSGroup.

SanikaGawhane: Moved `BOSH::Retryable` block inside the loop so that it retries on the single `ADD_MEMBERS` expression that failed in the current iteration.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -85,6 +85,11 @@ def to_s
     end
   end
 
+  class NSXTOptimisticUpdateError < StandardError
+    def initialize()
```
ktchen14, line 89: Doesn't seem like this is necessary. Also it's idiomatic in Ruby to leave off parentheses when calling or defining a function that takes no arguments.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -118,26 +123,32 @@ def initialize(client, default_vif_type)
...
-        grouping_obj_svc.add_or_remove_ns_group_expression(
-          nsgroup.id,
-          *to_simple_expressions(lports),
-          'ADD_MEMBERS'
-        )
+        Bosh::Retryable.new(
+            tries: 50,
+            on: [NSXTOptimisticUpdateError]
+        ).retryer do |i|
+          begin
```
ktchen14, line 135: You don't need the `begin` here. Since Ruby 2.5 `do ... end` blocks have an implicit `begin ... end`. See https://blog.bigbinary.com/2017/10/24/ruby-2.5-allows-rescue-inside-do-end-blocks.html.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -118,26 +123,32 @@ def initialize(client, default_vif_type)
...
+            grouping_obj_svc.add_or_remove_ns_group_expression(
+              nsgroup.id,
+              *to_simple_expressions(lports),
+              'ADD_MEMBERS'
+            )
+          rescue NSXT::ApiCallError => e
+            raise NSXTOptimisticUpdateError if e.code == 409 || e.code == 412 #Conflict or PreconditionFailed
+            raise e
```
ktchen14, line 143: Just `raise`. From the documentation "With no arguments, raises the exception in $! or raises a RuntimeError if $! is nil".

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -146,14 +157,23 @@ def remove_vm_from_nsgroups(vm)
...
-        grouping_obj_svc.add_or_remove_ns_group_expression(
-          nsgroup.id,
-          *to_simple_expressions(lports),
-          'REMOVE_MEMBERS'
-        )
+        Bosh::Retryable.new(
+            tries: 50,
+            on: [NSXTOptimisticUpdateError]
+        ).retryer do |i|
+          begin
```
ktchen14, line 166: Same comments as above.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
+          nsxt_provider.add_vm_to_nsgroups(vm, ns_groups)
+        end
+        it 'should raise an error if there is error adding vm to the nsgroup' do
+          failure_response = NSXT::ApiCallError.new(:code => 400)
+          expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_raise(failure_response)
+          expect do
+            nsxt_provider.add_vm_to_nsgroups(vm, ns_groups)
+          end.to raise_error(NSXT::ApiCallError)
+        end
+        it "should retry when there is a CONFLICT in adding vm to the nsgroup" do
```
ktchen14, line 212: Blank line above here.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
+          conflict_response = NSXT::ApiCallError.new(:code => 409, :response_body => 'The object was modified by somebody else')
```
ktchen14, line 213: Use new style hashes for symbol keys:

```ruby
conflict_response = NSXT::ApiCallError.new(code: 409, response_body: 'The object was modified by somebody else')
```

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
+          expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_return(true)
+          2.times{ expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_raise(conflict_response) }
```
ktchen14, line 215: I think you just want `.twice` here.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
+          expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_return(true)
           nsxt_provider.add_vm_to_nsgroups(vm, ns_groups)
         end
+        it "should retry when there is a PreconditionFailed in adding vm to the nsgroup" do
```
ktchen14, line 219: Blank line above here.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
+          precondition_failed_response = NSXT::ApiCallError.new(:code => 412, :response_body => 'PreconditionFailed')
+          4.times{ expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_raise(precondition_failed_response) } #retry 4 times
+          expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_return(true).twice
+          nsxt_provider.add_vm_to_nsgroups(vm, ns_groups)
+        end
+        it "should retry 50 times when there is a conflict or precondition failed while adding vm to nsgroup" do
```
ktchen14, line 225: Separate examples with a blank line so that they're differentiable on sight.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
+          conflict_response = NSXT::ApiCallError.new(:code => 409, :response_body => 'The object was modified by somebody else')
+          retryable = Bosh::Retryable.new()
```
ktchen14, line 227: No parentheses here.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
           nsxt_provider.add_vm_to_nsgroups(vm, ns_groups)
         end
+        it "should retry when there is a PreconditionFailed in adding vm to the nsgroup" do
+          precondition_failed_response = NSXT::ApiCallError.new(:code => 412, :response_body => 'PreconditionFailed')
+          4.times{ expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_raise(precondition_failed_response) } #retry 4 times
+          expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_return(true).twice
+          nsxt_provider.add_vm_to_nsgroups(vm, ns_groups)
+        end
+        it "should retry 50 times when there is a conflict or precondition failed while adding vm to nsgroup" do
+          conflict_response = NSXT::ApiCallError.new(:code => 409, :response_body => 'The object was modified by somebody else')
```
ktchen14, line 226: New style hashes.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
+          retryable = Bosh::Retryable.new()
+          allow(Bosh::Retryable)
```
ktchen14, line 228: Instead of this you might want an `allow_any_instance_of`. Can't decide if that's better. See https://relishapp.com/rspec/rspec-mocks/v/3-9/docs/working-with-legacy-code/any-instance.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -194,13 +195,45 @@
...
+              .to receive(:new)
+                      .with(hash_including(tries: 50))
+                      .and_return(retryable)
+          expect(grouping_obj_svc).to receive(:add_or_remove_ns_group_expression).with(any_args).and_raise(conflict_response)
+          expect do
```
ktchen14, line 233: Separate the actual test from the mocks with a blank line. There's a lot of setup and it's hard to understand the actual expectation.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
SanikaGawhane: `allow_any_instance_of` is indeed a better (and cleaner) way of writing tests. Modified the test accordingly.

PR #267, opened 2020-05-28
Check if PortGroup is NSXT type to support NSXT 3 on VDS

EleanorRigby:


src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
```diff
@@ -482,8 +482,22 @@ def find_child_by_name(mob, child_path)
...
+      end
+      return false if portgroup.nil?
+      return false unless portgroup.config.respond_to?(:backing_type)
+      logger.info("DVPG #{key} is backed by NSXT") if portgroup.config.backing_type == NSX_BACKING_TYPE
+      portgroup.config.backing_type == NSX_BACKING_TYPE
+    end
+
     private
 
+    NSX_BACKING_TYPE = 'nsx'.freeze
```
ktchen14, line 499: No need for this constant.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
```diff
@@ -482,8 +482,22 @@ def find_child_by_name(mob, child_path)
       find_child_by_name(mob.child_entity.find {|c| c.name == child_entity_name }, child_path)
     end
 
+    def dvpg_istype_nsxt?(key:, dc:)
+      logger.info("Checking if #{key} is backed by NSXT")
+      portgroup = dc.mob.network.detect do |network|
```
ktchen14, line 487: This looks initially good but we have to make this logic more complicated to account for the fact that the relevant portgroup could be nested inside something else. As far as I know this only looks at the direct children of the datacenter's `networkFolder`.

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
```diff
@@ -93,16 +96,17 @@
...
     end
   end
 
   describe '#set_vif_type' do
     let(:success_response) { HTTP::Message.new_response('') }
-
     before do
       allow(nsxt_provider).to receive(:logical_ports).with(vm).and_return([logical_port_1])
+      allow(nsxt_provider).to receive(:nsxt_nics).with(vm).and_return(nic)
+      allow(nic).to receive(:empty).and_return(false)
```
ktchen14, line 109: Hmm what's this?

SanikaGawhane: Every method checks `nsxt_nics(vm).empty?` even before attempting to find logical ports and moving ahead. I tried to re-use this pre-condition check using the `before` block for all the tests within the block.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
SanikaGawhane: I think it recursively checks all children of networkFolder. (verified using VC's mob browser). @EleanorRigby - can you please confirm if this is right?

src/vsphere_cpi/spec/unit/cloud/vsphere/nsxt_provider_spec.rb
ktchen14: No need to do this since `nic` is already an `Array` (actually it should probably be called `nics`) and will respond to `empty?` correctly. Also this line doesn't actually do anything right now. I think this was intended:

```ruby
allow(nic).to receive(:empty?).and_return(false)
```

Note the question mark. But either way it's unnecessary.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
EleanorRigby: I think Network will list all possible Networks in Datacenter. And using networkFolder to recursively gather should bring no difference.

EleanorRigby: We already use this in find_network to reliably get a list of all networks available under the Datacenter.

src/vsphere_cpi/lib/cloud/vsphere/nsxt_provider.rb
```diff
@@ -393,12 +395,21 @@ def to_simple_expressions(lports)
     end
 
     def nsxt_nics(vm)
-      nics = vm.nics.select do |nic|
-        nic.backing.is_a?(VimSdk::Vim::Vm::Device::VirtualEthernetCard::OpaqueNetworkBackingInfo) &&
-          nic.backing.opaque_network_type == NSXT_LOGICAL_SWITCH
+      nics = []
+      vm.nics.select do |nic|
```
ktchen14, line 399: I think this should be `vm.nics.each` rather than `vm.nics.select`. Overall I feel like this function is a little complicated. Here's my take:

```ruby
nics = vm.nics.select do |nic|
  case nic.backing
  when VimSdk::Vim::Vm::Device::VirtualEthernetCard::OpaqueNetworkBackingInfo
    next unless nic.backing.opaque_network_type == NSXT_LOGICAL_SWITCH
    logger.info("NIC with device key #{nic.key} is backed by opaque NSXT network")
  when VimSdk::Vim::Vm::Device::VirtualEthernetCard::DistributedVirtualPortBackingInfo
    next unless @vcenter_client.dvpg_istype_nsxt?(
      key: nic.backing.port.portgroup_key, dc: @datacenter
    )
    logger.info("NIC with device key #{nic.key} is backed by NSXT DVPG")
  end
  true
end
logger.info("NSX-T networks found for vm '#{vm.cid}': #{nics.map(&:device_info).map(&:summary)}")
nics
```

src/vsphere_cpi/lib/cloud/vsphere/soap_stub.rb
```diff
@@ -11,7 +11,12 @@ def initialize(vcenter_api_uri, http_client)
     end
 
     def create
-      base_stub = VimSdk::Soap::StubAdapter.new(@vcenter_api_uri, 'vim.version.version12', @http_client)
+      if $vc_version == "7.0"
```
ktchen14, line 14: Let's extract the assignment to `version` outside:

```ruby
      version = if $vc_version == "7.0"
        'vim.version.v7_0'
      else
        'vim.version.version12'
      end
```

src/vsphere_cpi/spec/unit/cloud/vsphere/soap_stub_spec.rb
```diff
@@ -8,21 +8,27 @@
...
-        )
-        .and_return(base_adapter)
-      expect(VSphereCloud::SdkHelpers::RetryableStubAdapter).to receive(:new)
-       .with(
-         base_adapter,
-       )
-       .and_return(retryable_adapter)
-
-      expect(soap_stub.create).to eq(retryable_adapter)
+    before(:all) do
```
ktchen14, line 11: Should be `before(:each)` and `after(:each)` probably. Also this is a perfect time to use `around` rather than `before` and `after`.

src/vsphere_cpi/spec/unit/cloud/vsphere/soap_stub_spec.rb
```diff
@@ -8,21 +8,27 @@
...
+      @old_vc_version = $vc_version
+    end
+    after(:all) do
+      $vc_version = @old_vc_version
+    end
+    context 'when vc_version is set to 7.0' do
```
ktchen14, line 17: Can we put blank lines between `before`, `after`, `context`s, and `it`s? It's hard to see all the different tests when there aren't blank lines.

src/vsphere_cpi/spec/unit/cloud/vsphere/soap_stub_spec.rb
```diff
@@ -8,21 +8,27 @@
...
+      before(:each) { $vc_version = '7.0' }
+      it 'returns the SDK Soap Adapter with VIM version set to vim.version.v7_0' do
+        expect(VimSdk::Soap::StubAdapter).to receive(:new).with('https://some-host/sdk/vimService', 'vim.version.v7_0', http_client,).and_return(base_adapter)
```
ktchen14, line 20: No need for a trailing comma here and on subsequent lines in the parentheses around a method call unless the arguments are split on separate lines. Though given how long this line is, we should probably split some stuff onto separate lines.

src/vsphere_cpi/spec/unit/cloud/vsphere/soap_stub_spec.rb
EleanorRigby: I will use `around` and fix spacing in the test. I am changin VC version in the tests, so I want to preserve and set the old value **once** for `all` tests

PR #268, opened 2020-06-03
Nsxt cvds 2

ktchen14:
This just refactors `find_network` to make it easier for us to implement resolution of ambiguous DVPG names.


src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,93 @@
...
+module VSphereCloud
+  module Resources
+    class Network
+      include VimSdk
+      include ObjectStringifier
+      stringify_with :name, :mob, :bosh_name
+
+      attr_reader :mob, :client, :bosh_name, :vim_name
+
+      def self.NetworkBuilder(bosh_name, mob, client)
```
ktchen14, line 12: I think this can just be called `make` or something like that.

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,93 @@
...
+        if mob.is_a?(VimSdk::Vim::Dvs::DistributedVirtualPortgroup)
```
ktchen14, line 13: `case ... when`

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,93 @@
...
+          OpaqueNetwork.new(bosh_name, mob, client)
+        else
+          Network.new(bosh_name, mob, client)
+        end
+      end
+
+      def initialize(bosh_name, mob, client)
+        @bosh_name = bosh_name
+        @mob = mob
+        @vim_name = @mob.name
```
ktchen14, line 25: Seems like this should be a method rather than an instance variable and `attr_reader`.

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,93 @@
...
+        backing_info
+      end
+
+      def inspect
+        "<network: / #{@bosh_name} / #{@mob}"
+      end
+    end
+
+    class OpaqueNetwork < Network
+      def initialize(*args, &block)
```
ktchen14, line 42: Unnecessary if the implementation is just `super`.

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -671,7 +671,21 @@ def generate_network_env(devices, networks, dvs_index)
         if device.kind_of?(Vim::Vm::Device::VirtualEthernetCard)
           backing = device.backing
           if backing.kind_of?(Vim::Vm::Device::VirtualEthernetCard::DistributedVirtualPortBackingInfo)
-            v_network_name = dvs_index[backing.port.portgroup_key]
+            network = @datacenter.mob.network.detect do |n|
```
ktchen14, line 674: Can we split this into a series of `#select` and `#detect`s to make it easier to read?

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,98 @@
...
+module VSphereCloud
+  module Resources
+    class Network
+      include VimSdk
+      include ObjectStringifier
+      stringify_with :name, :mob, :bosh_name
+
+      attr_reader :mob, :client, :bosh_name
+      def self.make_network_resource(bosh_name, mob, client)
+        case mob
```
ktchen14, line 12: I think the `case` statement should return the type. That way we don't repeat the `.new(bosh_name, mob, client)` clause.

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,98 @@
...
+
+      def nic_backing(_)
+        backing_info = VimSdk::Vim::Vm::Device::VirtualEthernetCard::NetworkBackingInfo.new
+        backing_info.device_name = vim_name
+        backing_info.network = mob
+        backing_info
+      end
+
+      def inspect
+        "<network: / #{@bosh_name} / #{@mob}"
```
ktchen14, line 44: I think you're missing a `>`

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+    end
+
+    class OpaqueNetwork < Network
+
```
ktchen14, line 49: Remove this extra blank line.

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,98 @@
...
+      def nic_backing(dvs_index)
```
ktchen14, line 50: Can this `dvs_index` be removed now that each network is an actual object? Previously networks weren't represented as actual objects so it was difficult to introspect them to build this index. Seems like it might be possible now?

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,98 @@
...
+        network_id = mob.summary.opaque_network_id
+        backing_info.opaque_network_id = network_id
+        dvs_index[network_id] = bosh_name
+        backing_info.opaque_network_type = mob.summary.opaque_network_type
+        backing_info
+      end
+    end
+
+    class DistributedVirtualPortGroupNetwork < Network
+
```
ktchen14, line 61: Extra blank line.

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,98 @@
...
+      def nic_backing(dvs_index)
+        portgroup_properties = client.cloud_searcher.get_properties(mob,
+                                                             VimSdk::Vim::Dvs::DistributedVirtualPortgroup,
```
ktchen14, line 64: Dedent this rather than aligning with the `(`. This is super hard to read.

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,98 @@
...
+        backing_info.port = port
+
+        dvs_index[port.portgroup_key] = bosh_name
+        backing_info
+      end
+    end
+
+
+    class DistributedVirtualPortGroupNSXTNetwork < Network
+
```
ktchen14, line 84: Extra blank line here.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
```diff
@@ -217,54 +217,69 @@ def find_network_retryably(datacenter, network_name)
...
+      return nil if networks.empty?
+      return networks.first if networks.length == 1
+      # If it is VC 7.0, then NSXT 3.0 can configure same Logical switch as separate dvpgs
+      # under each VC cluster's VDS. But all such dvpg map back to same LS in NSXT.
+      # Hence, disambiguation is needed
+      # Check if all portgroups are actually same by checking these conditions
+      # 1. Network type is DVPG
+      # 2. Network responds to backing type (or check vc 7.0) and Network has nsx backing type
+      # 3. Logical Switch ID for DVPG exists && it is same for all (Same as first newtwork in list)
+      network = networks.first if networks.all? do |n|
```
ktchen14, line 270: I think you actually have to do `return network.first if ...` because if you don't then it might get overridden on the next line.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
```diff
@@ -217,54 +217,69 @@ def find_network_retryably(datacenter, network_name)
...
+        n.is_a?(VimSdk::Vim::Dvs::DistributedVirtualPortgroup) &&
```
ktchen14, line 271: Yeah I think turning this into a series of `#select`s followed by a single `#all` will make it easier to read.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
EleanorRigby: Only gets overwritten if it is nil. So safe

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: done

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: done

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
ktchen14: Yeah it still makes more sense to just return the network outright here and remove the `if network.nil?` on the next line.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
EleanorRigby: We cannot select and then apply all?  because all of these networks must adhere to conditions. If there is a network that is not meeting conditions, we do not want to deselect and ignore but we want all? to return false.

src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
EleanorRigby: This can be split.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
ktchen14: Oh you're totally right.

src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
EleanorRigby: got it

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: Done

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
```diff
@@ -0,0 +1,110 @@
...
+module VSphereCloud
+  module Resources
+    class Network
+      include VimSdk
+      include ObjectStringifier
+      stringify_with :name, :mob, :bosh_name
+
+      attr_reader :mob, :client, :bosh_name
+      def self.make_network_resource(bosh_name, mob, client)
```
ktchen14, line 11: I don't think we need `bosh_name` at all if we're not populating `dvs_index`. Instead of `#get_dvs_index_hash`, can we make a function like `#index_name` that returns the index key that should go in such a hash? That way this network doesn't need to know about the BOSH name whatsoever.

src/vsphere_cpi/spec/integration/nsxt_spec.rb
```diff
@@ -87,8 +87,7 @@
     }
   end
   let(:network_spec) do
-    {
-      'static-bridged' => {
+    spec = {'static-bridged' => {
```
ktchen14, line 90: Is this change necessary?

EleanorRigby: nope, will remove it

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/network_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+    describe '#get_dvs_index_hash' do
+      it 'returns empty hash' do
+        dvs_index = subject.get_dvs_index_hash
+        expect(dvs_index).to eq({})
+      end
+    end
+  end
+
+  describe '#self.make_network_resource' do
+    before(:each) do
```
ktchen14, line 38: This can be done in the `let` itself.

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/network_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+  describe '#vim_name' do
+    it 'should return network with vim name as fake-network' do
+      expect(subject.vim_name).to eq(vim_network_name)
+    end
+  end
+
+  describe '#nic_backing' do
+    it 'returns backing of type plain NetworkBackingInfo ' do
+      backing_info = subject.nic_backing
+      expect(backing_info).to be_a(VimSdk::Vim::Vm::Device::VirtualEthernetCard::NetworkBackingInfo)
```
ktchen14, line 25: `be_an_instance_of` to make sure that it's not a subclass.

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: for now, yes we can make that change. 

But being a network resource object of CPI  I think it is fine to put the variable in that class too

ktchen14: Actually maybe `#network_id` is a better name?

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/network_spec.rb
EleanorRigby: done

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/network_spec.rb
EleanorRigby: This is not for `instance_double` but for actual vim definition

src/vsphere_cpi/lib/cloud/vsphere/resources/network.rb
EleanorRigby: Done

PR #269, opened 2020-06-08
Clean up find_network integration tests

ktchen14:


src/vsphere_cpi/spec/integration/client_spec.rb
```diff
@@ -38,105 +38,119 @@ module VSphereCloud
...
-          network = @client.find_network(@datacenter, "#{folder_name}/non-existent-standard-portgroup")
-          expect(network).to be_nil
+        it 'returns the standard switch if it exists' do
+          network = @client.find_network(@datacenter, standard_portgroup_path)
+          expect(network).to be_an_instance_of(VimSdk::Vim::Network)
+          expect(network.name).to eq(standard_portgroup_path.split('/').last)
+        end
+
+        it 'returns the DVPG if it exists' do
+          network = @client.find_network(@datacenter, dvpg_path)
```
EleanorRigby, line 142: This and test below test with same `dvpg_path`
```
dvpg_path is = BOSH_VSPHERE_CPI_PORTGROUP_FOLDER_DISTRIBUTED=vcpi-network-folder/vcpi-switch-2/vcpi-dvpg-2-1
```

I think this test should just use `vcpi-network-folder/vcpi-dvpg-2-1` (without DVS)

ci/config.rb
```diff
@@ -1,7 +1,8 @@
 $pipeline.pool('6.0-NSXV') do |pool|
   pool.params = {
     RSPEC_FLAGS: [
-      '--tag ~nsx_transformers',
+      '--tag ~nvds',
+      '--tag ~cvds',
```
EleanorRigby, line 5: We can run `vsphere_networking` here too?

ktchen14: Hmm it wasn't running before. I'm incline to say let's leave enabling `vsphere_networking` here for later?

PR #303, opened 2021-09-10
Initial commit to convert fully to cpi version 2

jpalermo:
- Fixes a problem where attaching persistent disk can fail because
  the agent cannot reload the cdrom config in time

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -6,7 +6,8 @@ require 'cloud'
 require 'cloud/vsphere/cpi_extension'
 
 module VSphereCloud
-  class Cloud < Bosh::Cloud
+  class Cloud
+    include Bosh::CloudV2
     include VimSdk
     include Logger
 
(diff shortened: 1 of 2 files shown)


src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
```diff
@@ -644,6 +645,7 @@ def attach_disk(vm_cid, raw_director_disk_cid)
         # Overwrite cid with the director cid
         # Since director sends messages with "director cid" to agent, the agent needs that ID in its env, not the clean_cid
         add_disk_to_agent_env(vm, director_disk_cid, disk_spec.device.unit_number)
```
jpalermo, line 647: `add_disk_to_agent_env` updates the vm config bosh in the datastore and on the cdrom and then reattaches it to the VM. Ideally it wouldn't need to do this.

The cdrom should not need to be reattached at all as part of this change, but both of those actions are handled by the same function. I'm not clear if keeping the `env.json` in the datastore up to date is important or not. Perhaps the `env.json` is only needed in the datastore to recreate the cdrom for this particular use case, and persisting that data is no longer needed. I'm not familiar with all the flows through the CPI, so I left this untouched, but you should take a look.

julian-hj:
@cunnie & I merged this manually and added tests. It is rolling through our pipelines now, and🤞 we should have results by tomorrow morning. We will cut a new release if all goes well.

PR #320, opened 2022-04-18
Adding support for ruby 3.1.x

lnguyen:


ramonskie:
im currently getting a warning
```
ARNING: An expectation of `:exitstatus` was set on `nil`. To allow expectations on `nil` and suppress this message, set `RSpec::Mocks.configuration.allow_message_expectations_on_nil` to `true`. To disallow expectations on `nil`, set `RSpec::Mocks.configuration.allow_message_expectations_on_nil` to `false`.
```
this could be disabled by adding `mocks.allow_message_expectations_on_nil = true`
on line 57 in the spec_helper.rb
https://github.com/cloudfoundry/bosh-vsphere-cpi-release/blob/master/src/vsphere_cpi/spec/spec_helper.rb#L57

src/vsphere_cpi/.bundle/config
```diff
@@ -0,0 +1,4 @@
+---
+BUNDLE_CACHE_ALL: "false"
+BUNDLE_NO_INSTALL: "true"
+BUNDLE_WITHOUT: "development:test"
```
ramonskie, line 4: this will cause issues running tests
as those gems are not installed

ramonskie (approved):

cunnie (approved):

rkoster:
Thanks @lnguyen

PR #329, opened 2022-06-14
Refresh group before each call to add_vm_to_group() when using nsxt policy provider

### LINKED ISSUE #328: Bosh sometimes silently fails to add a vm to a group when using nsxt policy provider
**Describe the bug**

We've noticed when using nsxt policy api that bosh can think the vm has been successfully added to a group when it has not.

When modifying group members with the policy api, each request to update the group contains a revision id. If multiple requests to nsxt are made to modify the same group with the same revision id, only one of the requests will succeed. The others will receive a 4xx code, and the cpi will log a message in the task with "Revision Error: ..." and retry.

I think whats going on is in the first attempt in `add_vm_to_group()`, the vm is added to `group_obj.expression`, and then nsxt is called to modify the group, which can fail if the same group is being modified for a different vm such that each request has the same revision id.
In the second attempt, the code uses the same `group_obj` from the first attempt, and sees the vm is already in the `group_obj` expression. `add_vm_to_group()` then returns early without having made a successful api call.

https://github.com/cloudfoundry/bosh-vsphere-cpi-release/blob/v74/src/vsphere_cpi/lib/cloud/vsphere/nsxt_policy_provider.rb#L284-L287
```
        existing_external_id = external_id_expressions.find { |expr| expr.external_ids.include?(vm_external_id) }
        if existing_external_id != nil
          return
        end
```

The bosh task will then continue and complete successfully without error.

To fix, after a revision error and before retrying, the cpi could make a new request to nsxt for an updated copy of the group before trying to modify it again.

**To Reproduce**

The issue requires bosh to be adding more than one vm to the same group at the same time to hit a revision error in the nsxt policy api. We first noticed the problem when concourse was applying a cloud-config change to 5 deployments in parallel, where the director was recreating a vm from each and adding them to the same group at close to the same time. Our director has 12 workers.

1. Create 5 deployments, each with an instance_group with vm_extension to the same group:

deployment1:
```
instance_groups:
- name: foo
  vm_extensions: [mygroup_extension]
...
```

deployment2:
```
instance_groups:
- name: bar
  vm_extensions: [mygroup_extension]
...
```

2. Ensure the cloud config vm extension applied to each instance has the same nsxt nsgroup:
```
vm_extensions:
- name: mygroup_extension
  cloud_properties:
    nsxt:
      ns_groups: [mygroup]
```

3. Start an operation which will involve creating these vms at similar times. We first hit this as our concourse starts 5x tasks which do a `bosh deploy` to different deployments in parallel, and we were deploying a change to cloud-config which meant all vms needed to be recreated. I was also able to reliably reproduce the problem using a bash loop and backgrounding each recreate e.g.
```bash
for N in $(seq 1 5); do
  bosh recreate -n -d deployment${N} &
done
```

4. Wait for the 5 tasks to finish with State "done", meaning bosh thinks they were successful and no errors occurred.

5. Check each task cpi log for "Revision Error" e.g. `bosh task 123 --cpi | grep "Revision Error"`. The presence of this will indicate that multiple requests were made to nsxt with the same revision id, such that at least one failed and was retried, meaning the bug was hit.

6. Check the "mygroup" members in nsxt. At least one of the vms which were recreated will not be a member of "mygroup".

**CPI Error Log**

The tasks which have hit the bug will have something like this:

> I, [2022-06-05T23:30:40.866607 #20075]  INFO -- [req_id cpi-533437]: Adding vm: foo_deployment1_f1bcb6eb9028 to groups: ["mygroup"]
> I, [2022-06-05T23:30:41.131129 #20075]  INFO -- [req_id cpi-533437]: Adding vm foo_deployment1_f1bcb6eb9028 to group {:_revision=>2, :_system_owned=>false, :display_name=>"mygroup", :description=>"", ...
> I, [2022-06-05T23:30:41.625775 #20075]  INFO -- [req_id cpi-533437]: Revision Error: while adding vm: foo_deployment1_f1bcb6eb9028 to group mygroup with message The object was modified by somebody else. Please retry.
> I, [2022-06-05T23:30:42.384757 #20075]  INFO -- [req_id cpi-533437]: Finished create_vm in 22.82 seconds

**Expected behavior**

After a revision error is encountered, the cpi should retry adding the vm to the group, or eventually error if it does not succeed. The bosh task would then show with State = error, making it more obvious that the vm was not added to a group.

**Release Version & Related Info (please complete the following information):**

 - CPI Version v73
 - BOSH Director Version 272.2.0
 - Stemcell Name & Version ubuntu-bionic 1.87
 - vCenter Version 7.x
 - NSXT Version 3.x

**Additional context**

We first noticed this with vsphere-cpi v73, and from a quick check of the code I believe this is still present in v74, but I'll re-test after we've updated.

It might be possible to also reproduce this within a single deployment if `update` is set such that multiple instances are recreated at the same time. But I think that'd mean they are within the same director worker, and Im not sure if you could still get a Revision Error from nsxt.

If we made fewer concurrent bosh operations we probably wouldnt have seen this, but due to the size of our environment and number of bosh deployments we've parallelised these calls to increase the speed of rolling out updates.

bg-govau:
# Description

A quick fix adding a refresh of the group object each time before caling add_vm_to_group().

## Related PR and Issues
Fixes #328 

- nsxt policy provider

## Type of change

## How Has This Been Tested?

We've deployed and tested the change using the process to reproduce from the issue - i.e. ran `bosh recreate` on multiple deployments which attempt to add vms to the same security group. 

**Test Configuration**:
vCenter Version 7.x
NSXT Version 3.x

# Checklist:

- [x] ~I have added tests that prove my fix is effective or that my feature works~
- [x] ~New and existing unit tests pass locally with my changes~

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/nsxt_policy_provider.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/nsxt_policy_provider.rb
@@ -60,6 +60,7 @@ module VSphereCloud
 
       groups.each do |group|
         retry_on_conflict("while adding vm: #{vm.cid} to group #{group.display_name}") do
+          group = retrieve_group(group_id: group.id)
           add_vm_to_group(group, vm.cid)
         end
       end


ae-govau:
Would anyone be available to review @bg-govau 's change?  It's only a one-liner, and mirrors the existing pattern used for `remove_vm_from_groups()` a few lines lower, and fixes a really dangerous race-condition described in #328 (where BOSH think it's succeeded, but has actually failed).

ragaskar:
This looks good to me. I believe we actually removed this line because we did not understand its significance. It would be nice to have a unit test or something to prevent this kind of accident from happening again, but in the meantime this feels safe to merge. 

I think we may also need to do this with server pools.

jpalermo:
Thanks!

ragaskar:
Hi @bg-govau  just wanted to see if you've had a chance to try against HEAD and if this solved your problem? 

Thanks!

bg-govau:
Hi @ragaskar, no I was just waiting for the next release to be cut. As an interim solution we've been applying the same fix in a pre-start script to modify nsxt_policy_provider.rb, and it has fixed the problem for us, so we're pretty happy its ok and havent bothered building our own release.

bg-govau:
Thanks for the release, confirming we havent seen the problem since upgrading to it and removing our pre-start workaround.

PR #339, opened 2022-10-12
Expose memory_reservation_locked_to_max via cloud_properties

rkoster:
# Description
By setting this property all guest memory will be reserved. If a VM has more memory reserved than is available in a given resource pool it will fail to power on.

For multi-cluster configurations where resource pool memory reservations have been set some logic has been added which will try the next available cluster in case of a VM power-on failure.

## Type of change

## How Has This Been Tested?
Integration tests have been successfully run against one of the pooled ci environments.

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/vm_type.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/vm_type.rb
@@ -15,7 +15,7 @@ module VSphereCloud
     end
 
     %w[cpu ram disk cpu_hot_add_enabled memory_hot_add_enabled nsx vmx_options
-       nsxt nested_hardware_virtualization upgrade_hw_version datacenters vm_group disable_drs storage_policy tags].each do |name|
+       nsxt memory_reservation_locked_to_max nested_hardware_virtualization upgrade_hw_version datacenters vm_group disable_drs storage_policy tags].each do |name|
       define_method(name) { @cloud_properties[name] }
     end
 
--- a/src/vsphere_cpi/spec/integration/multi_cluster_overflow_spec.rb
+++ b/src/vsphere_cpi/spec/integration/multi_cluster_overflow_spec.rb
@@ -99,7 +99,7 @@ context 'given 2 clusters with a datastore and a resource pool' do
       update_resource_pool_memory_reservation(cpi, @first_cluster_name, @first_rp, 1, true)
       update_resource_pool_memory_reservation(cpi, @second_cluster_name, @second_rp, 1, true)
     end
-    it 'should create a VM on the first cluster' do
+    it 'should create a VM on the second cluster' do
       begin
         @vm_id, _ = cpi.create_vm(
           'agent-007',
--- a/src/vsphere_cpi/spec/integration/multi_cluster_overflow_spec.rb
+++ b/src/vsphere_cpi/spec/integration/multi_cluster_overflow_spec.rb
@@ -1,0 +1,123 @@
+require 'integration/spec_helper'
+
+context 'given 2 clusters with a datastore and a resource pool' do
+  before (:all) do
+    @first_cluster_name = fetch_and_verify_cluster('BOSH_VSPHERE_CPI_CLUSTER')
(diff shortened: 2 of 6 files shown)


src/vsphere_cpi/lib/cloud/vsphere/vm_type.rb
```diff
@@ -15,7 +15,7 @@ def initialize(datacenter, cloud_properties, pbm)
     end
 
     %w[cpu ram disk cpu_hot_add_enabled memory_hot_add_enabled nsx vmx_options
-       nsxt nested_hardware_virtualization upgrade_hw_version datacenters vm_group disable_drs storage_policy tags].each do |name|
+       nsxt memory_reservation_locked_to_max nested_hardware_virtualization upgrade_hw_version datacenters vm_group disable_drs storage_policy tags].each do |name|
```
cunnie, line 18: Looks good.
I _believe_ that by setting this to true & attempting to power on the VM & if there's not enough memory (`cluster_free_memory` - `overall_usage`) it won't power on.

We should, at some point, update the bosh.io docs to reflect this new property.

From the vSphere docs:

> If set true, memory resource reservation for this virtual machine will always be equal to the virtual machine's memory size; increases in memory size will be rejected when a corresponding reservation increase is not possible. This feature may only be enabled if it is currently possible to reserve all of the virtual machine's memory

src/vsphere_cpi/spec/integration/multi_cluster_overflow_spec.rb
```diff
@@ -0,0 +1,123 @@
+require 'integration/spec_helper'
```
cunnie, line 1: I **love** this integration test.

cunnie (approved):
Nice job!

cunnie (commented):
I'd amend the commit message slightly to show the use case (_why_ someone would use this), e.g.:

> Setting memory_reservation_locked_to_max: true might prove helpful when excessive VM migration is seen, evidenced by kernel messages such as "CPU#2 stuck for 23s!". This can happen when vSphere HA is enabled with admission control on clusters with constrained memory resources.

cunnie (approved):
LGTM

nouseforaname (changes requested):
Other than the description of the 2nd test case, lgtm

src/vsphere_cpi/lib/cloud/vsphere/vm_type.rb
rkoster: Wat ahead of you 🙃 https://github.com/cloudfoundry/docs-bosh/pull/779

src/vsphere_cpi/spec/integration/multi_cluster_overflow_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+  context 'when the second cluster has enough reservable memory' do
+    before do
+      update_resource_pool_memory_reservation(cpi, @first_cluster_name, @first_rp, 100, false)
+      update_resource_pool_memory_reservation(cpi, @second_cluster_name, @second_rp, 1024, false)
+    end
+    after do
+      update_resource_pool_memory_reservation(cpi, @first_cluster_name, @first_rp, 1, true)
+      update_resource_pool_memory_reservation(cpi, @second_cluster_name, @second_rp, 1, true)
+    end
+    it 'should create a VM on the first cluster' do
```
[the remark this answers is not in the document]
rkoster: Fixed

nouseforaname (approved):
thanks!

src/vsphere_cpi/spec/integration/multi_cluster_overflow_spec.rb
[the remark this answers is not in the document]
aramprice: Seems so

aramprice (approved):
Looks good to me. Seems like there's one spec description that should be updated at
> src/vsphere_cpi/spec/integration/multi_cluster_overflow_spec.rb:102

`first cluster` => `second cluster`

PR #340, opened 2022-10-13
Make try next cluster on power on failure work with vSphere 6.5

rkoster:
The changes introduced in: baf997dd1b323015e93dc91b512e396249cf3708 did not work on vSphere 6.5 because the exception raised was of a different type.


lnguyen (approved):
LGTM

cunnie (approved):
LGTM

PR #341, opened 2022-10-13
In NSX-T Policy mode, expect `nsxt[ns_groups]` to specify Group ids.

ragaskar:
# Description
- Groups and NS Groups can be retrieved from the NSX-T API by ID or display name. Policy API Groups are an equivalent to Management API NS Groups.
- Group IDs and display_names are not guaranteed to match.
  - On the Policy API side, we've found that for UI-created Groups, the ID _will_ match the display_name _if_ 1) the display name does not contain special characters (like spaces, slashes) AND 2) the display name is unique.
  - On the Management API side, a NSGroup created through the UI appears to have a GUID generated as the ID (it will _never_ match the display_name).
- When talking to the Management API the CPI uses a NS Group display_names to retrieve groups.
- After this commit, when talking to the Policy API, the CPI uses Group ids to retrieve groups.
- We've _deliberately_ chosen to retrieve Groups by ID from the Policy API, even though this does not match Management API behavior.
- This represents a return to the behavior used when the Policy API code was originally written. This effectively reverts 6113c106df6c7a9164b9b76f465143224b8dda0c.
- This resolves a problem for Policy API users who have Groups with non-unique or invalid-as-an-id display_names.
- However, it introduces a potential problem for users when in migration mode. Migration mode intends to update Groups in both Management and Policy APIs, and after this change, the NS Group display name in the Management API _must match_ the Group id in the Policy API for the migration mode to work as expected.
- The net effect of this change is that it is the users' responsibility in a migration scenario to ensure that the passed `ns_groups` values will work correctly for each API.
- `NsxtPolicyProvider#retrieve_groups_by_name` was removed because it is no longer used.

NSX-T behavior

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)
- [-] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [X]  Integration tests
- [X] Unit tests

# Checklist:

- [X ] My code follows the standard ruby style guide
- [X ] I have performed a self-review of my own code
- [X ] I have commented my code, particularly in hard-to-understand areas
- [X ] I have added tests that prove my fix is effective or that my feature works
- [X ] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -939,10 +939,10 @@ module VSphereCloud
         @nsxt_policy_provider.add_vm_to_server_pools(created_vm, static_server_pools) if static_server_pools
       end
       groups = []
-      groups = @nsxt_policy_provider.retrieve_groups_by_name(ns_groups) unless ns_groups.empty?
+      groups = @nsxt_policy_provider.retrieve_groups_by_id(ns_groups) unless ns_groups.empty?
       if (ns_groups.count > groups.count)
         if allow_missing_resources
-          logger.info("Not all specified groups found, missing #{(ns_groups - groups.map(&:display_name)).join(",")}. VM will still be added to found groups (#{groups.map(&:display_name).join(",")})")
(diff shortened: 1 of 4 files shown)


cunnie (approved):
Looks GREAT! Nice job, Rajan.

cunnie (approved):
Really looks good!

rkoster (approved):

Issue #336, opened 2022-09-22
TAS deployment failure during network churn


suyash2796:
Stemcell: ubuntu-xenial/621.261
Bosh: 2.10.46-build.541
TAS 2.13.8

Issue description:

While we are trying to scale diego cell count in our TAS env and having network/org churn running in parallel, we are facing below issue due to which deployment is failing. This issue can also be observed during NSXT tile upgrade from ops manager and having network/org churn running in parallel.

```
Task 96 | 09:05:47 | Creating missing vms: diego_cell/a8a9036f-bc39-4412-b73a-0b920b3050de (14) (00:53:37)
Task 96 | 09:05:56 | Creating missing vms: diego_cell/cf1f94cd-790b-407b-bedb-eea685604974 (15) (00:53:46)
                  L Error: Unknown CPI error 'Unknown' with message 'The object 'vim.dvs.DistributedVirtualPortgroup:dvportgroup-22218' has already been deleted or has not been completely created' in 'set_vm_metadata' CPI method (CPI request ID: 'cpi-432182')
Task 96 | 09:05:56 | Creating missing vms: diego_cell/57211cee-682c-4023-b84e-77331e12ac5c (17) (00:53:46)
                  L Error: Unknown CPI error 'Unknown' with message 'The object 'vim.dvs.DistributedVirtualPortgroup:dvportgroup-22218' has already been deleted or has not been completely created' in 'set_vm_metadata' CPI method (CPI request ID: 'cpi-668394')
Task 96 | 11:25:18 | Creating missing vms: diego_cell/06d316fe-1269-4b41-bd2c-d48f180ed3fe (18) (03:13:08)
Task 96 | 11:28:38 | Creating missing vms: diego_cell/205ec15c-0873-4079-8d06-68df49bd8c00 (13) (03:16:28)
Task 96 | 11:30:39 | Creating missing vms: diego_cell/3e9a5f94-21be-470a-a8b8-de54b35486f8 (10) (03:18:29)
Task 96 | 11:31:57 | Creating missing vms: diego_cell/37d64fad-c101-4fc3-a2d1-bb378e7e85d4 (19) (03:19:47)
Task 96 | 11:31:57 | Error: Unknown CPI error 'Unknown' with message 'The object 'vim.dvs.DistributedVirtualPortgroup:dvportgroup-22218' has already been deleted or has not been completely created' in 'set_vm_metadata' CPI method (CPI request ID: 'cpi-432182')
[...]

PR #343, opened 2022-10-19
Attempt to prefetch portgroup data

klakin-pivotal:
This commit largely follows the pattern established by the fix in
<https://github.com/cloudfoundry/bosh-vsphere-cpi-release/pull/332>.
Given that the complaint in
<https://github.com/cloudfoundry/bosh-vsphere-cpi-release/issues/336>
is similar to the one that caused the fix in #332, I believe that following the same pattern may be helpful.

The hope is that `find_resources_by_property_path` will both only select `DistributedVirtualPortgroup` objects, and also pre-fetch the `key` property of all of the objects returned. If it does this, then we should avoid attempting to fetch information about deleted vSphere networking objects, and only get the class of objects we're interested in.

Do note that I've only run the unit tests (which are pretty thoroughly mocked out), so the integration tests may fail.

[#183583921] BOSH-172 - TAS deployment failure during network churn

Port Group fetching

## Type of change

## How Has This Been Tested?

- [x] Unit tests run locally

**Test Configuration**:
`be rspec spec` was run on a Bosh Ecosystem Team standard workstation.

# Checklist:

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/vcenter_client.rb
@@ -547,10 +547,14 @@ module VSphereCloud
 
     def dvpg_istype_nsxt?(key:, dc_mob:)
       logger.info("Checking if #{key} is backed by NSXT")
-      portgroup = dc_mob.network.detect do |network|
-        #network.key is same as network.config.key
-        network.is_a?(VimSdk::Vim::Dvs::DistributedVirtualPortgroup) && network.key == key
+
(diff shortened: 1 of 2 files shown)


cunnie (approved):
LGTM!

ragaskar:
I'd like to run this through integration tests before throwing a thumbs up (I'll take care of this bit). LMK if we need to move faster here.

nouseforaname (approved):
Nice, Thanks.

ragaskar:
Running integration tests now ...

ragaskar:
Most tests are passed (the nsxt3.1 is still running, but I expect it to go green), seems safe to merge.

PR #345, opened 2022-10-20
Increase ram to fix 6.5 tests

lnguyen:
The 6.5 test beds have some weirdness around setting ram to 100 in reservations. It is looking for at least 640 so update_resource_pool_memory_reservation was actually failing and it never tried to create vm and start it up. This bumps it up and it seems to work


rkoster (approved):

ragaskar:
this is unnecessary, I think. These failures occur because a VM gets "stuck" when the first test erroenously places a VM in the second cluster. At that point one can't change the mem "back" to less than 512 if expandable if false (since the VM running in that resource pool prevents it).

PR #350, opened 2022-11-01
Only fetch required DVPGs in generate_network_env #183413921

nouseforaname:
[ #183413921 ]

if we're running on vsphere >= v7 and nsx-t >= v3 and are using NSX-T on vsphere DVS

then

NSX-T will potentially sync thousands of NSX-T logical switches into a DVS in certain conditions (the setup where this was discovered used NCP on Kubernetes within the same VCENTER. NCP and VCENTER shared the NSX-T deployment. NCP created logical switches for every namespace on a cluster leading to thousands of synced DVPGs in VCENTER).

and that caused create_vm to run painfully long since the old code iterated over `datacenter.mob.networks` to find the right DVPG within all network objects in vsphere. Checking all items in the list could take enough time for some items in the list being deleted from NSX before they could be accessed.

We do not need to iterate over all networks since the `device.port.portgroup_key` got populated in create_vm which calls find_network_retryably to properly deal with identifying the right network to use. So it should be safe to use data on the previously configured device to search for just the DVPG we need and (depending on whether the DVPG is created via NSX) use either the nsx switch id or the portgroup_key for `v_network_name`

Update existing tests for extra cloudsearcher calls, add additional test for "synced" DVPGs.

Tests:
- units
- integration:
-- `bundle exec rspec spec/integration/ --tag ~nvds --tag cvds --tag nsxt_all --tag vsphere_networking` against `vsphere 7` + `nsx 3.1` with cvds networks configured
-- `bundle exec rspec --tag '~cvds' --tag '~disk_migration' --tag '~nsxv' --tag '~host_maintenance' --tag '~nsxt_policy'` against `vsphere 6.5` + `nsx 2.5`

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -777,21 +777,49 @@ module VSphereCloud
         next unless device.kind_of?(Vim::Vm::Device::VirtualEthernetCard)
         v_network_name = case device.backing
           when Vim::Vm::Device::VirtualEthernetCard::DistributedVirtualPortBackingInfo
-            network = @datacenter.mob.network.detect do |n|
-              n.is_a?(VimSdk::Vim::Dvs::DistributedVirtualPortgroup) &&
-                n.config.respond_to?(:backing_type) &&
-                n.config.backing_type == 'nsx' &&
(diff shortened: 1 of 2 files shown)


lnguyen (approved):
LGTM

cunnie (approved):
LGTM

PR #351, opened 2022-11-01
Use `ruby-3.1` package which includes Clang

cunnie:
Previously, the CPI used the old ruby-3.1.2-r0.91.0 package to avoid package collision with the BOSH Director. This had several drawbacks, especially that it was difficult to autobump and had poor performance on Jammy.

This commit addresses both shortcomings by switching to using the newer, more easily autobump-able ruby-3.1 package, which also includes the ability to build with the Clang compiler instead of the GCC compiler.

The Clang compiler generates a smaller, faster Ruby than GCC on Jammy stemcells.

Specifically, the thread memory footprint is much smaller, from our notes:

- Jammy w/ GCC: 21688 initial, 32156 after thread, 32156 after finish, around 1MB per Thread.new call
- Jammy w/ Clang: 13932 initial, 14012 after thread, 14208 after finish, between 8KB and 28KB per Thread.new call


lnguyen (approved):
LGTM

PR #353, opened 2022-12-05
expose vmx_options.disk.enableUUID globally

nouseforaname:
operators may choose to expose this setting globally instead of adding it to all vm_types in cloud config.

why:

As an operator who mainly deploys kubernetes clusters that use the kubernetes vsphere container storage plugin, I do not wan't to set the disk.enableUUID flag on a per vm_type basis. I want to enable this globally.

adds test for behaviour when the options is set on cpi and vm level. VM level settings take precedence over cpi level settings.

the config itself is optional and only rendered in if required.

### DIFF
--- a/jobs/vsphere_cpi/spec
+++ b/jobs/vsphere_cpi/spec
@@ -77,6 +77,9 @@ properties:
       -----BEGIN CERTIFICATE-----
       MII...
       -----END CERTIFICATE-----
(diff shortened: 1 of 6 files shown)


ramonskie (approved):
lgtm

lnguyen (approved):
LGTM

PR #357, opened 2023-03-01
X-Allow-Overwrite header is configurable. 

ragaskar:
# Description

When the policy API was introduced, the CPI began using the `X-Allow-Overwrite` header for all NSXT API requests. It's not entirely clear why this choice was made, in general this behavior is only to be used in exceptional scenarios (e.g., permissions problem with orphan data). 

Some users have seen failures when they target Policy API created resources but configure the CPI to use the management plane api. This configuration "works", as with the addition of the `X-Allow-Overwrite` header, the (normally protected) Policy API resources can be modified. However there are certain edge cases where this fails -- with this configuration, users have seen unexpected behavior during NSX-T upgrades that resulted in incorrect pool membership. 

Eventually we should remove the use of `X-Allow-Overwrite`, but this removal could suddenly break users who are (knowingly or not) relying on the existing behavior. Instead, we've made the use of this header configurable (defaulting to "true", so that users must opt-in to abandoning the header and the new behavior that comes with it). 

In instances where the CPI is using the NSX-T Management Plane API, we also log out when the CPI modifies a "PROTECTED" object; this may help users affected by the edge case failures scenarios described above more quickly identify root cause of their problem. 

* NSX-T API (no behavior change)

## Type of change
* Additional configuration

## How Has This Been Tested?

Ran existing unit + integration pipelines

# Checklist:

- [X ] My code follows the standard ruby style guide
- [ X] I have performed a self-review of my own code
- [ X] I have commented my code, particularly in hard-to-understand areas
- [ X] I have added tests that prove my fix is effective or that my feature works
- [ X] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/vsphere_cpi/data/swagger-nsxt-manager-template/api_client.mustache
+++ b/src/vsphere_cpi/data/swagger-nsxt-manager-template/api_client.mustache
@@ -26,8 +26,7 @@ module {{moduleName}}
       @user_agent = "{{#httpUserAgent}}{{{.}}}{{/httpUserAgent}}{{^httpUserAgent}}Swagger-Codegen/#{VERSION}/ruby{{/httpUserAgent}}"
       @default_headers = {
         'Content-Type' => 'application/json',
-        'User-Agent' => @user_agent,
-        'X-Allow-Overwrite' => true
+        'User-Agent' => @user_agent
       }
     end
 
@@ -87,6 +86,12 @@ module {{moduleName}}
         end
       end
       response
+    end
+
+    # Enables the deletion of test artifacts such as NS Groups and Logical Routers
+    # Don't enable in production code; it enables Management API to modify Policy objects
+    def x_allow_overwrite(value = true)
+      @default_headers['X-Allow-Overwrite'] = value
     end
 
     # Builds the HTTP request
--- a/src/vsphere_cpi/lib/cloud/vsphere/config.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/config.rb
@@ -1,5 +1,5 @@
 module VSphereCloud
-  class NSXTConfig < Struct.new(:host, :username, :password, :remote_auth, :auth_certificate, :auth_private_key, :default_vif_type, :use_policy_api, :policy_api_migration_mode)
+  class NSXTConfig < Struct.new(:host, :username, :password, :remote_auth, :auth_certificate, :auth_private_key, :default_vif_type, :use_policy_api, :policy_api_migration_mode, :allow_overwrite)
     def self.validate_schema(config)
       return true if config.nil?
 
@@ -28,6 +28,11 @@ module VSphereCloud
 
     def policy_api_migration_mode?
       !!policy_api_migration_mode
+    end
+
+    def allow_overwrite?
+      return true if allow_overwrite.nil?
+      allow_overwrite
     end
   end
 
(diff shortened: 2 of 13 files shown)


ragaskar:
Intent is to add docs to bosh.io post-cpi cut

rkoster:
@ystros & @danielfor could you both take a look at this PR?

ragaskar:
thanks for the nudge. I also added a couple of other folks :)

src/vsphere_cpi/lib/cloud/vsphere/config.rb
```diff
@@ -29,6 +29,10 @@ def use_policy_api?
     def policy_api_migration_mode?
       !!policy_api_migration_mode
     end
+
+    def allow_overwrite?
+      allow_overwrite || true
```
ystros, line 34: The `|| true` here will prevent anyone from overriding the default value. It will continue always setting the `X-Allow-Overwrite` header.

src/vsphere_cpi/spec/unit/cloud/vsphere/config_spec.rb
```diff
@@ -316,6 +316,39 @@ module VSphereCloud
...
+            'host' => 'fake-host',
+            'username' => 'fake-username',
+            'password' => 'fake-password',
+            'certificate' => nil,
+            'private_key' => nil,
+            'allow_overwrite' => false
+          }
+        end
+        it 'returns set value if false' do
+          expect(config.nsxt.policy_api_migration_mode?).to be(false)
```
ystros, line 347: It should assert on `allow_overwrite?` intsead of policy api migration mode.

src/vsphere_cpi/data/swagger-nsxt-manager-template/api_client.mustache
```diff
@@ -89,6 +88,12 @@ module {{moduleName}}
       response
     end
 
+    # Enables the deletion of test artifacts such as NS Groups and Logical Routers
+    # Don't enable in production code; it enables Management API to modify Policy objects
+    def x_allow_overwrite(value = true)
+      @default_headers['X-Allow-Overwrite'] = value
+    end
```
ystros, line 95: If I understand correctly, we don't have direct control over the `config` object passed to the ApiClient constructor, and thus need to call this separate method to update the headers. Is that correct?

ystros (changes requested):

src/vsphere_cpi/lib/cloud/vsphere/config.rb
ragaskar: nice, bad code allowed by a bad test. 🤦

src/vsphere_cpi/data/swagger-nsxt-manager-template/api_client.mustache
ragaskar: I think we could similarly override the Swagger generated configuration object (like we are elsewhere) or otherwise add an extra param to the NSXT API client (which is swagger generated, but which we already template). 

Initially it seemed like this functionality was more of a rush, and so I chose not to do either of these things. 

Long-term I'm not excited about having a lot of mutations in the swagger-generated parts with the anticipation that this is an additional maintenance burden. I think trade-off makes both changes equally unappealing and this one seemed a little faster (and also kind of preserves a separation between "our" config and the "NSXT API Client" config, which is already a little confusing). 

I dunno, we could make this more "seamless" but I can't yet come up with any really good reasons to do so.

ragaskar:
Corrections have been force-pushed.

jpalermo (approved):
👍

ystros (approved):

PR #359, opened 2023-03-09
[WIP]Feature/#184562415 retry create vm on fallback datastore if create vm fails with a full datastore

nouseforaname:
vsphere has two (or more datastores) that run at high capacity we can
end up in a situation where the datastore that is prioritized via the
weighting provided in the selection pipeline cannot fit ALL VMs that
need to be created. That wouldn't be a problem if we wouldn't run these
in threads. Since the threads are not aware about each other and do the
same weighting, both threads, when running at the same time, will pick
the same datastore.

Though if that datastore can only fit one more VM, then one of the
threads will raise the newly introduced error. To avoid erroring a bosh
deployment that would otherwise succeed, we can retry the create_vm on
the next datastore from the selection pipeline if we encounter that
start_vm fails with GenericVmConfigFault. To properly handle this error,
this introduces and update to the retry judge. We did internally retry this
error when encountered since it wasn't int the list. That internal retry
to power_on added additional runtime to the deployment but didn't
solve anything.

To my understanding this behaviour should be specific to usage of
multiple datastores (as opposed to the cpi using a datastore cluster) so
no change was added to the codepaths dealing with clusters. additionally
this whole behaviour only applies to the DS chosen for the ephemeral
disk. The error message for a failed deployment was:

```
Error running task 'Drm.ExecuteVmPowerOnLRO'. Failed with message 'Module MonitorLoop power on failed. ' and fault mes
sage 'Module MonitorLoop power on failed. ,Failed to extend swap file /vmfs/volumes/<GUID>/vm-<GUID>/vm-<GUID>.vswp
from 0 KB to 67108864 KB : No space left on device,Current swap file size is 0 KB.,
Failed to extend the virtual machine swap file,Could not power on virtual machine:
No space left on device.,Failed to power on VM.,Failed to start the virtual machine.'.
```

This scenario is now retried under the condition that the selection
pipeline contained other possible datastores that were less appropriate
at time of check. To avoid restarting the checks we just cache the
fallback datastores in the newly introduced attribute of VmPlacement.


selzoc (commented):
This looks generally OK to me.

I did notice some files with changed behavior, (`vm_placement_selection_pipeline.rb` for example) that do not have corresponding test file changes.

lnguyen:
@selzoc @nouseforaname has added tests for `vm_placement_selection_pipeline.rb` do you want to look over this?

selzoc:
LGTM 👍

PR #361, opened 2023-04-02
Faster deploys by trimming NSX auto-generated objects

cunnie:
BOSH vSphere CPI is slower when NSX is configured, increasing deploy times by as much as 10%. Specifically, the CPI call [`set_vm_metadata`](https://bosh.io/docs//cpi-api-v2-method/set-vm-metadata/) can double the amount of time to complete (e.g. from 5 seconds to 10 seconds).

We can mitigate this by only using the APIs we need, for earlier investigation showed that much of the additional time was spent loading the thousands of NSX classes.

This commit makes the following reductions:

- NSX Policy
  - Endpoints (e.g. `/infra/lb-pools`): 1299 → 8
  - Models/Ojbects (e.g. `SegmentPort`): 1344 → 75
- NSX Management (deprecated)
  - Endpoints (e.g. `/infra/lb-pools`): 872 → 24
  - Models/Ojbects (e.g. `SegmentPort`): 1260 → 146

We reduced the total number of non-test Ruby files in the vSphere CPI 3049 → 348, which represents an 88% reduction in the number of files.

Notes on endpoint changes:

- `read_lb_pool_0()` → `read_lb_pool()` because the original `read_lb_pool`, generated from the endpoint `/global-infra/lb-pools/{lb-pool-id}`, is no longer instantiated. We don't use that endpoint.
- Similarly with `update_lb_pool_0()`.
- Similarly, `get_tier1_segment_port_0()` → `get_tier1_segment_port` due to unused endpoint `/global-infra/tier-1s/{tier-1-id}/segments/{segment-id}/ports/{port-id}`
- And `patch_tier1_segment_port_0`, too.

The screenshot below shows CPI v88, CPI v90, CPI v90 w/ trimmed NSX, and CPI v90 with full NSX. Note that the first three CPIs take approximately the same time to deploy and delete 2 deployments of 10 VMs apiece, whereas the full NSX CPI takes 10% longer:

Screenshot 2023-04-02 at 7 21 15 AM

# Description

This makes the vSphere CPI faster when NSX is configured. Deploying new VMs can be up to 10% faster

## Related PR and Issues

_No related PRs or GitHub Issues_

This affects NSX, both Management API and Policy API.

This does not affect NSX-V.

## Type of change

Performance improvement.

## How Has This Been Tested?

It's been tested against the following set of lifecycle tests:

- [X] lifecycle-6.5-nsxt25
- [X] lifecycle-6.7-nsxt25
- [X] lifecycle-7.0-nsxt30
- [X] lifecycle-7.0-nsxt30
- [X] lifecycle-7.0-nsxt31
- [X] lifecycle-7.0-nsxt31-cvds

**Test Configuration**:

The following parameters were copied from the Production lifecycle tests:

* Toolchain:
* SDK:

# Checklist:

- [x] (N/A) I have added tests that prove my fix is effective or that my feature works


jpalermo (approved):
While I admit I did not review all 2,724 changed files (especially as the github UI seems to suffer greatly with the diff); the concept and execution seem very sane and I'd trust our tests to notice any missing files.

PR #363, opened 2023-04-19
Add timeouts to config for Typhoeus to prevent NSX requests hanging for hours

jpalermo:
New timeout is 10 minutes

Default Typhoeus connect timeout is 5 minutes, so we left that alone.


rkoster (approved):

ramonskie (commented):
i think it would be nice if all the files containing the  swagger codegen version changes.
in a separate commit.
as it would be easier to navigate and see whats the actuall changes are

cunnie:
> i think it would be nice if all the files containing the swagger codegen version changes. in a separate commit. as it would be easier to navigate and see whats the actuall changes are

Done!

cunnie (approved):
LGTM

PR #364, opened 2023-04-19
Remove blobstore info from IaaS metadata endpoints

ystros:
The BOSH agent reads its blobstore settings from the metadata API endpoint (or equivalent) for its VM within the IaaS. If the blobstore settings are not set in the `env.bosh.blobstores` property, it will fallback to the top-level `blobstore` property in the metadata.

However, in modern configurations, the Director always sends the blobstore settings as part of the environment hash. Additionally, the Director does redaction of credentials in the environment hash when the signed URLs blobstore feature is enabled. This redaction is not applied to the top-level `blobstore` property in the metadata because that is generated solely by the CPI.

Rather than updating each CPI to know about the signed URL feature, we are instead removing the `blobstore` properties from the CPI. This will ensure that Director is the sole point of contact when configuring agent blobstore settings, and ensure that they are always properly redacted.

### DIFF
--- a/jobs/vsphere_cpi/spec
+++ b/jobs/vsphere_cpi/spec
@@ -99,71 +99,10 @@ properties:
     default:
       - 0.pool.ntp.org
       - 1.pool.ntp.org
-  agent.blobstore.access_key_id:
-    description: AWS access_key_id for agent used by s3 blobstore plugin
-  agent.blobstore.secret_access_key:
-    description: AWS secret_access_key for agent used by s3 blobstore plugin
   agent.mbus:
     description: Agent mbus
   agent.nats.address:
     description: Address of the nats server
-  agent.blobstore.address:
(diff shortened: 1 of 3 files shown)


ystros:
It should be noted that this is potentially a breaking change for users who are not already sending their blobstore settings via the `env` hash. At the very least, `bosh-deployment` is already properly sending it for both the [Director settings (provider=local)](https://github.com/cloudfoundry/bosh-deployment/blob/f0ecc13d3a8f3c0b7d6666ede368591277376759/bosh.yml#L160-L163) and for [VMs deployed by the director](https://github.com/cloudfoundry/bosh-deployment/blob/f0ecc13d3a8f3c0b7d6666ede368591277376759/bosh.yml#L38-L46)

rkoster (approved):

aramprice (approved):

PR #365, opened 2023-06-10
Allow datastore clusters to be nested in folders

jpalermo:
Currently datastore clusters must be top level for the cpi to find them, this change will add functionality to search for them within storage folders in vcenter.

Slashes are not valid in folder or datastore cluster names, they must be escaped in vCenter, so we use slash as our expected folder delimiter.

For the "pattern" matching case, after splitting on slashes, we have to convert each segment into a regular expression in case the user is relying on that for wildcard matching. As regexes don't use forward slashes, and the entity names they are searching for also can't have slashes, this should be safe.

The unused StoragePod.find method was also removed

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/storage_pod.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/storage_pod.rb
@@ -52,18 +52,30 @@ module VSphereCloud
       end
 
       def self.search_storage_pods(name_pattern, datacenter_mob)
-        path_segments = name_pattern.split('/')
-        current_folders = [datacenter_mob.datastore_folder]
-        while path_segments.length > 1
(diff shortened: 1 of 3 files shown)


lnguyen (approved):
LGTM

jpalermo:
We realized that this behavior will not work if you specify a datastore pattern that has multiple regex options already. i.e. `/folder1/datastore1|/folder2/datastore2|datastore3`

jpalermo:
We've updated `search_in_subfolders` to now build a full index of datastore clusters (StoragePods) and the folders they are in, then use the provided pattern to match the appropriate ones.

cunnie (approved):
LGTM

rkoster:
@lnguyen could you take an other look at the changes?

lnguyen (approved):
LGTM

PR #368, opened 2023-07-03
Fix TLS verification with multiple CPI configs

ystros:
Previously, the CPI wrote the CA certs for vCenter, NSX, and NSX-T to disk as part of BOSH templating. This works correctly for environments with a single CPI config. However, in environments that have 2 or more vSphere CPI configs, the CA certs for the specific config used are passed at runtime by the BOSH Director when invoking the CPI. The CPI was always using the CA certs rendered to disk, meaning the CAs in other CPI configs would be ignored and could cause TLS verification failures during the deploy.

Now, instead of rendering the CA certs to disk during BOSH templating, they are written to disk as temporary files once the updated config for the specific CPI is received and merged (the `context` variable in the lambda in the `vsphere_cpi` executable). Temporary files were chosen rather than re-writing the templated file to avoid problems when multiple CPI invocations occur concurrently.

### DIFF
--- a/jobs/vsphere_cpi/templates/cpi.erb
+++ b/jobs/vsphere_cpi/templates/cpi.erb
@@ -20,9 +20,6 @@ export NO_PROXY="<%= no_proxy %>"
 export no_proxy="<%= no_proxy %>"
 <% end %>
 
-<% if_p('vcenter.connection_options.ca_cert') do %>
-  export BOSH_CA_CERT_FILE=$BOSH_JOBS_DIR/vsphere_cpi/config/cacert.pem
-<% end %>
 <% if_p('vcenter.nsx.ca_cert') do %>
   export BOSH_NSX_CA_CERT_FILE=$BOSH_JOBS_DIR/vsphere_cpi/config/nsx_cacert.pem
 <% end %>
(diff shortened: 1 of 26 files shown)


jobs/vsphere_cpi/templates/cpi.erb
```diff
@@ -20,15 +20,7 @@ export NO_PROXY="<%= no_proxy %>"
...
-  export BOSH_CA_CERT_FILE=$BOSH_JOBS_DIR/vsphere_cpi/config/cacert.pem
-<% end %>
-<% if_p('vcenter.nsx.ca_cert') do %>
-  export BOSH_NSX_CA_CERT_FILE=$BOSH_JOBS_DIR/vsphere_cpi/config/nsx_cacert.pem
-<% end %>
-<% if p('vcenter.nsxt.ca_cert', false) %>
-  export BOSH_NSXT_CA_CERT_FILE=$BOSH_JOBS_DIR/vsphere_cpi/config/nsxt_cacert.pem
-<% else %>
+<% if !p('vcenter.nsxt.ca_cert', false) %>
   export NSXT_SKIP_SSL_VERIFY=true
```
jpalermo, line 24: Is NSXT_SKIP_SSL_VERIFY not also something that should be determined by the config passed by the director?

ystros: Good catch, we're running the integration tests with this environment variable removed. Will resolve this comment once they go green.

ystros: Tests went green w/ the removal.

jpalermo (approved):
👍

PR #377, opened 2024-01-26
Enable expanding root disk size with new property `root_disk_size_gb`

cunnie:
Applications such as Nvidia's libraries, required for many AI / LLM applications, aren't easily installed on anywhere but the root disk; however, the amount of free space in the root disk of a Jammy stemcell is too small to accommodate these libraries.

To address this, we introduce a new property, `root_disk_size_gb`. Setting this property will expand the root disk during VM creation (in the `create_vm` CPI method, before the BOSH packages are installed).

Typical use (in a Cloud Config):

```yaml
vm_extensions:
- name: 20G_root
  cloud_properties:
    root_disk_size_gb: 20
```

Setting this property disables the VM's "linked clone" feature [0], which is incompatible with extending the size of the root disk. The VM will therefore require more disk space than expected, typically the size of the root disk (in the above example, 20 GiB)

Adding this feature brings the vSphere CPI to parity with the AWS, Azure, and GCP CPIs.

Root disk size is measured in GiB (1,073,741,824 bytes), not in GB (1,000,000,000 bytes).

Drive-by:

Removed unnecessary initialization, `config_spec.device_change = []`: `ConfigSpec` class initializes `device_change` to an empty array; we don't need to do it.

[0] A linked clone is a copy of a virtual machine that shares virtual disks with the parent virtual machine. Only changes from the parent disk are recorded, saving on disk space.

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Related PR and Issues
Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [X] Includes an integration test which was run on vSphere 7.0

# Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
@@ -76,6 +76,10 @@ module VSphereCloud
       vm_type.pci_passthroughs || []
     end
 
+    def root_disk_size_gb
+      vm_type.root_disk_size_gb.to_i
+    end
+
     def storage_policy_name
       @manifest_params[:storage_policy]
     end
--- a/src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
@@ -183,6 +183,8 @@ module VSphereCloud
 
           # Clone VM
           logger.info("Cloning vm: #{replicated_stemcell_vm} to #{vm_config.name}")
+          # Don't link clone if expanding root disk, otherwise "Invalid operation for device '0'. Disks with parents cannot be expanded."
+          linked = vm_config.root_disk_size_gb > 0 ? false : true
           created_vm_mob = @client.wait_for_task do
             @cpi.clone_vm(
               replicated_stemcell_vm.mob,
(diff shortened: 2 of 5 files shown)


src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -76,6 +76,10 @@ def pci_passthroughs
       vm_type.pci_passthroughs || []
     end
 
+    def root_disk_size_gb
+      vm_type.root_disk_size_gb.to_i
```
jpalermo, line 80: Does this blow up if `root_disk_size_gb` isn't specified?

src/vsphere_cpi/lib/cloud/vsphere/vm_creator.rb
```diff
@@ -261,7 +262,6 @@ def create(vm_config)
             end
             unless vm_config.pci_passthroughs.empty?
               config_spec = VimSdk::Vim::Vm::ConfigSpec.new
-              config_spec.device_change = []
```
jpalermo, line 264: Is there any concern with this no longer being set to anything in some cases? Previously this was set to empty array, but now it only gets configured to something if root_disk_size_gb > 0.

If you don't specify root_disk_size_gb, this won't get set to empty array. Is that okay?

cunnie: > Is there any concern with this no longer being set to anything in some cases?

Actually, it's *always* set to an empty array. When you create a new `VimSdk::Vim::Vm::ConfigSpec` (which is `config_spec`'s class), it initializes the `device_change` accessor to an empty array. There's no reason to set it to an empty array a second time.

jpalermo (approved):
👍

PR #381, opened 2024-06-24
Fix lifecycle-8.pre-release-nsxt41-cvds CI job.

gm2552:
# Description

The lifecycle-8.pre-release-nsxt41-cvds job started failing on April 11th due to a change in the platform that it is connecting to.  Specifically, the certificate associated with the NSXT Manager changed it's TLS certificate to use a wild card cert and the hard coded NSXT Manager host name will no long work (fails with SSL errors when trying to connect to the NSX Manager APIs.

This change dynamically creates an NSXT Manager host name based on information in the SAN of the TLS server certificate.  It is aware of SANs that use specific host names or wildcard domain names.  In the case of a wildcard, a new FQDN is derived from the wild card cert's domain name.  This hostname added to the /etc/hosts file and used as the NSX Manager hostname throughout the tests.

## Related PR and Issues
NA

CI only.

## Type of change

## How Has This Been Tested?

Tested in a temporary CI Pipeline that had accessed to all necessary infrastructure (Nimbus pools) that pointed to a forked repo for the CI task updates (the same forked repo that is part of this PR).  Successfully ran the lifecycle-8.pre-release-nsxt41-cvds job from that temporary pipeline.

To reproduce, updated the pipeline.yaml to point the `source-ci` Git resource to the repo referenced in this PR and install the pipeline on a Concourse instance that has access to the Nimbus pools (can currently only runs on VMware hosted pipelines).  Run the lifecycle-8.pre-release-nsxt41-cvd job and validate that all tests pass.

### DIFF
--- a/ci/tasks/run-lifecycle.sh
+++ b/ci/tasks/run-lifecycle.sh
@@ -37,7 +37,7 @@ if [ "$BOSH_VSPHERE_CPI_NSXT_HOST" != "null" ]; then
   openssl s_client -showcerts -connect "${BOSH_VSPHERE_CPI_NSXT_HOST}":443 </dev/null 2>/dev/null | openssl x509 -outform PEM > $BOSH_NSXT_CA_CERT_FILE
   # The certificate's SAN contains the host name, so extract it because SSL validation fails when using the IP address
   # NOTE: Don't hard code the name as it is not guaranteed to be "nsxt-manager"
-  BOSH_NSXT_CERT_HOST_NAME="$(openssl x509 -noout -text -in nsxt-manager-cert.pem | awk '/X509v3 Subject Alternative Name/ {getline;gsub(/ /, "", $0); print}' | tr -d "DNS:")"
+  BOSH_NSXT_CERT_HOST_NAME="$(openssl x509 -noout -text -in nsxt-manager-cert.pem | awk '/X509v3 Subject Alternative Name/ {getline;gsub(/ /, "", $0); print}' | tr -d "DNS:" | sed 's/^\*.//')"
   echo "${BOSH_VSPHERE_CPI_NSXT_HOST} ${BOSH_NSXT_CERT_HOST_NAME}" >> /etc/hosts
   export BOSH_VSPHERE_CPI_NSXT_HOST="https://${BOSH_NSXT_CERT_HOST_NAME}"
 fi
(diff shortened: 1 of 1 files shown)


ci/tasks/run-lifecycle.sh
```diff
@@ -35,9 +35,11 @@ if [ "$BOSH_VSPHERE_CPI_NSXT_HOST" != "null" ]; then
   BOSH_NSXT_CA_CERT_FILE="$(realpath "${PWD}/nsxt-manager-cert.pem")"
   # To get the cert from nsxt-manager, we run openssl on the jump box, and then pipe that result into a local openssl command that reformats it into PEM
   openssl s_client -showcerts -connect "${BOSH_VSPHERE_CPI_NSXT_HOST}":443 </dev/null 2>/dev/null | openssl x509 -outform PEM > $BOSH_NSXT_CA_CERT_FILE
-  # The certificate's subject name is nsxt-manager so SSL validation fails when using the IP address
-  echo "${BOSH_VSPHERE_CPI_NSXT_HOST} nsxt-manager" >> /etc/hosts
-  export BOSH_VSPHERE_CPI_NSXT_HOST="https://nsxt-manager"
+  # The certificate's SAN contains the host name, so extract it because SSL validation fails when using the IP address
+  # NOTE: Don't hard code the name as it is not guaranteed to be "nsxt-manager"
+  BOSH_NSXT_CERT_HOST_NAME="$(openssl x509 -noout -text -in nsxt-manager-cert.pem | awk '/X509v3 Subject Alternative Name/ {getline;gsub(/ /, "", $0); print}' | tr -d "DNS:" | sed 's/^\*./nsx-mgr\./')"
```
mingxiao, line 40: > sed 's/^\*./nsx-mgr\./

the `.` on the search string is not escaped, so it will match anything. Perhaps escape just to be safe?

in the replacement string, you've escaped the dot, `\.`. We don't believe that escape is necessary here.

To remove some of the text process, you could use `openssl x509 -noout -in nsxt-manager-cert.pem -ext 'subjectAltName'`. The output of that command is something like:
```
X509v3 Subject Alternative Name:
    DNS:github.com, DNS:www.github.com
```

gm2552: Added the suggestions of escaping the . in the search string and removing the escape in the replacement string.

Adding `-ext 'subjectAltName` is resulting in the new name being returned twice, so I'm leaving out that suggestion for now.

mingxiao (approved):

PR #382, opened 2024-06-28
NSX: tags VMs as well as ports

klakin-pivotal:
# Description

This commit updates the 'set_vm_metadata' operation to tag NSX VM objects with the same set of tags that are being set on vSphere VM objects. This commit also adds a configuration option to enable to disable this new tagging: 'vcenter.nsxt.tag_nsx_vm_objects'. This new tagging is disabled by default. Also added is the generated code required to access the endpoint used in this new feature.

Why:

For large users with automation, NSX segment port tagging is not enough: Specifically, when setting membership criteria of Groups, the tag matching operators are too limited:

- Tag: Equals, Not Equals, Not In
- Scope: Equals, Not Equals

This prevents writing generic rules such as, "if the Scope is 'bosh/deployment' and the Tag begins with 'cf-' and the Scope is 'bosh/instance_group' and the Tag is 'router', then place it in the Web group"

But tagging the VMs offers a richer set of criteria:

- Tag: Equals, Contains, Starts With, Ends With
- Scope: Equals

This is relevant for one particular user that wants to match against the beginning of our 'bosh/deployment' tags. They cannot do an 'Equals' match because the deployment name has an unpredictable final component. This requires a 'Starts With' matcher to be useful.

Gotchas:

It appears that the endpoint we're using to tag the VMs only exists in the Manager API, not in the Policy API. Because there is no equivalent of this endpoint in the Policy API, we're fairly confident that this is one of the APIs which just happens to live in the "management" API namespace, and that this change will continue to work with the upcoming NSX 5.0 which is removing certain parts of the Management API.

## Related PR and Issues
N/A

NSX tagging, CPI configuration, additional work in the `set_vm_metadata` call.

## Type of change

## How Has This Been Tested?

- [X] Manual testing against a live 4.x NSX-T
- [X] Automated testing against NSX 3.0 with the integration tests included 

**Test Configuration**:
* Tests run were configured in the environments in the Concourse Jobs set up by the `ci/configure.sh` script.

# Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/jobs/vsphere_cpi/spec
+++ b/jobs/vsphere_cpi/spec
@@ -88,6 +88,9 @@ properties:
   vcenter.nsxt.allow_overwrite:
      description: "When enabled, the CPI sets the X-Allow-Overwrite header to 'true' when making NSX-T API requests. This value currently defaults to true for backwards compatibility."
      default: true
+  vcenter.nsxt.tag_nsx_vm_objects:
+     description: "When enabled, tag NSX VM objects with the same set of tags as vsphere VM objects"
+     default: false
   vcenter.http_logging:
     description: Enables HTTP level logging. Each HTTP request to vcenter will be logged
     default: false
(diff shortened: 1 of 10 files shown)


jpalermo (approved):
👍

PR #281, opened 2020-11-17
New VM type cloud properties for CPU and Memory reservation

svrc:
# Description

Latency sensitive VMs require CPU and Memory reservations.  This commit adds two new VM-level cloud properties:
* `memory_reservation_locked_to_max` - this translates directly into the [memoryReservationLockedToMax VIM API property](https://vdc-download.vmware.com/vmwb-repository/dcr-public/98d63b35-d822-47fe-a87a-ddefd469df06/2e3c7b58-f2bd-486e-8bb1-a75eb0640bee/doc/vim.vm.ConfigInfo.html) 
* `cpu_reserve_full_mhz` - this translates into [cpuAllocation VIM API property](https://vdc-download.vmware.com/vmwb-repository/dcr-public/98d63b35-d822-47fe-a87a-ddefd469df06/2e3c7b58-f2bd-486e-8bb1-a75eb0640bee/doc/vim.vm.ConfigInfo.html) by querying the [ESXi host's hardware summary](https://vdc-download.vmware.com/vmwb-repository/dcr-public/3325c370-b58c-4799-99ff-58ae3baac1bd/45789cc5-aba1-48bc-a320-5e35142b50af/doc/vim.host.Summary.HardwareSummary.html) across the designated Host Group or Cluster, and multiplying the VM's requested CPU count by the hardware CPU base clock MHz. 

## Related PR and Issues
fixes #4

## Impacted Areas in Application
These are optional properties and should have no impact on existing deployments.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## How Has This Been Tested?

Unit Tests TBD.

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

# Checklist:

- [X] My code follows the standard ruby style guide
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #384, opened 2024-07-15
New VM type cloud property for CPU reservation

rkoster:
Rebase on original PR failed: https://github.com/cloudfoundry/bosh-vsphere-cpi-release/pull/281 so opening new PR here.

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
@@ -14,7 +14,7 @@ module VSphereCloud
       CLUSTER_VM_HOST_RULE_SHOULD = 'SHOULD'.freeze
       CLUSTER_VM_HOST_RULE_MUST = 'MUST'.freeze
       PROPERTIES = %w(name datastore resourcePool host)
-      HOST_PROPERTIES = %w(name hardware.memorySize runtime.connectionState runtime.inMaintenanceMode runtime.powerState)
+      HOST_PROPERTIES = %w(name hardware.memorySize runtime.connectionState runtime.inMaintenanceMode runtime.powerState summary.hardware.cpuMhz)
       HOST_COUNTERS = %w(mem.usage.average)
 
       MEMORY_HEADROOM = 128
--- a/src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
@@ -8,6 +8,7 @@ module VSphereCloud
     def initialize(manifest_params:, cluster_provider: nil)
       @manifest_params = manifest_params
       @cluster_provider = cluster_provider
+      @max_mhz = 0
     end
 
     def upgrade_hw_version?(vmtype_hw_version, global_hw_version)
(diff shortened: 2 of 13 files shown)


ystros:
Opted against adding a full integration spec, since the feature takes a multiple of the ESXi host CPU MHz, which wouldn't be very predictable depending on the environment the tests are run.

We did manually deploy a VM using a VM extension to set `cpu_reserve_full_mhz: true` and saw in vCenter that the VM had reserve CPU equal to the ESXi host it was deployed on. All other VMs had 0 MHz reservation.
Screenshot 2024-07-23 at 12 18 16 PMScreenshot 2024-07-23 at 12 18 35 PM

ystros:
Corresponding docs PR: https://github.com/cloudfoundry/docs-bosh/pull/846

dsboulder:
I wish it was a % of CPU to reserve, rather than just 0 or 100. Like I might choose to reserve 1 pCPU (which is in terms of physical ghz) for every 2 vCPUs to let the hyperthreading do its thing.

dsboulder:
Also what about taking the MIN(cpufreq) across the cluster rather than the largest node? Or we think they're the same usually?

rkoster:
@svrc as the author of the original PR, do you happen to remember the reasoning behind the current logic?

jpalermo:
@dsboulder & @rkoster doesn't seem like we're going to get a response on this. What direction do you want to go with it?

src/vsphere_cpi/lib/cloud/vsphere/resources/cluster.rb
```diff
@@ -14,7 +14,7 @@ class Cluster
       CLUSTER_VM_HOST_RULE_SHOULD = 'SHOULD'.freeze
       CLUSTER_VM_HOST_RULE_MUST = 'MUST'.freeze
       PROPERTIES = %w(name datastore resourcePool host)
-      HOST_PROPERTIES = %w(name hardware.memorySize runtime.connectionState runtime.inMaintenanceMode runtime.powerState)
+      HOST_PROPERTIES = %w(name hardware.memorySize runtime.connectionState runtime.inMaintenanceMode runtime.powerState summary.hardware.cpuMhz)
```
jpalermo, line 17: HOST_PROPERTIES doesn't seem like it's used anymore and can be deleted rather than updated

src/vsphere_cpi/lib/cloud/vsphere/vm_config.rb
```diff
@@ -35,6 +36,18 @@ def name
...
+        @maxMhz = 0
+        cluster.host.each do |host|
+          if host.runtime.connection_state == 'connected' &&
+              !host.runtime.in_maintenance_mode
+                @maxMhz = host.summary.hardware.cpu_mhz if host.summary.hardware.cpu_mhz > @maxMhz
+          end
+        end
+      end
+    end
+
```
jpalermo, line 50: So we're finding the fastest cpu of any host in the cluster and saving that off. Then later, we're multiplying that by the number of CPUs in the VM type to get the actual "cpu_allocation" reservation.

This feels a bit odd. Shouldn't we just do the multiplication here of `host.summary.hardware.cpu_mhz` with `host.summary.hardware. num_cpu_cores` and then use that value later.

Not totally sure it matters in the end though. Is it a requirement that your "vm_type" number of CPUs matches exactly the number of cores in the host to be able to use this feature at all?

jpalermo (changes requested):

jpalermo (approved):
During FIWG discussion, it was decided that the feature "reserves the number of mhz * the number of cpus your vm type asks for" and the number of mhz is based on the largest host in the cluster you are targeting.

svrc:
Sorry for not responding earlier,  IIRC the logic for using the minimum CPU in the cluster is that you want to allow the workload to fit on any host in the cluster, not the largest host.   If you pick the largest host it will only be scheduled there.

PR #386, opened 2025-01-07
Handle non-JSON responses from NSX API endpoints

ystros:
Previously, if an NSX API endpoint returned a non-successful status code with a non-JSON response, the CPI would raise a parse error that was hard to decypher. Now, it will raise the expected error class with information about the error, include the body and status code, like other API errors.


aramprice (approved):

klakin-pivotal (approved):
Looks good to me.

PR #388, opened 2025-07-25
Fix `disk.enableUUID`

Alphasite:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

This pr is primarily intended to resolve some issues with the `disk.enableUUID` flag which doesn't operate correctly in many cases. 

To properly test it and backfill missing tests i also ended up making a refactor to the cloud/vmcreator to move the agent env methods. These are stacked as 2 separate commits in the same which can be reviewed independently but are difficult to split into entirely separate PRs as one is necessary for the other and the tests wont pass independently without further refactoring.

I've included the body of the 2 commits as they try to explain the changes.

> Refactor vmcreator.rb to move generate agent env from cloud.rb
> 
> This code is only used by the vm creator and creates a rather unpleasant cyclical dependency that is awkward to test and splits disk uuid handling across multiple classes. So i moved it into vm creator which linearises the call tree. Its no longer: `Cloud.create_vm -> VMCreator.create -> Cloud.generate_*_env`, it is now: `Cloud.create_vm -> VMCreator.create -> VMCreator.generate_*_env`
> 
> [TNZ-46888]

and

> Fix inconsistent/incorrect handling of disk.enableUUID between global and vm type config
> 
> The global cpi config flag to enable uuids does not work for several reasons:
>  1. It is not propagated into the vm extraConfig so will not be used for all disks.
>  2. The expected value string value is inconsistent across sites, and some expect it to be `'1'`, '1' or 'TRUE' depending on the author which makes it impossible to enable the feature correctly in all cases.
> 
> There are probably a couple more, but they escape me.
> 
> [TNZ-46888]

This PR is also stacked on top of the pvscsi and hardware version pr, but will block on their merging before ask for a full review of this pr.

## Related PR and Issues
Fixes TNZ-46888

- cloud.rb - This primarly affects the create_vm and attach_disk cpi calls
- vmcreator.rb

## Type of change

## How Has This Been Tested?

- [ ] New unit tests
- [ ] Manual deployment (needs some further testing)

**Test Configuration**:
* Toolchain:
* SDK:

# Checklist:

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/cloud.rb
@@ -366,7 +366,8 @@ module VSphereCloud
             global_clusters: @datacenter.clusters,
             disk_configurations: disk_configs,
             storage_policy: policy_name,
-            enable_human_readable_name: config.human_readable_name_enabled?
+            enable_human_readable_name: config.human_readable_name_enabled?,
+            enable_disk_uuid: config.disk_uuid_is_enabled?
           }
 
           if config.human_readable_name_enabled?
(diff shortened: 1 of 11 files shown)


julian-hj:
We started talking about this in gchat, but I had a minor doubt about what will happen if we turn on disk.enableUUID by default everywhere.

From my experimentation with TKR, it appears to be important for the vmx_options on the k8s worker vm to get `disk.enableUUID=TRUE` in order for the CSI driver on the node to be able to identify disks that were attached to the VM to fulfill PVCs. But I ran across [this thread](https://community.broadcom.com/vmware-cloud-foundation/discussion/why-is-diskenableuuidtrue-not-the-default#:~:text=If%20a%20VM%20with%20disk%20UUIDs%20enabled%20is%20used%20as,nor%20about%20the%20quiesce%20messages.) which suggests that we might run into trouble if the same setting is applied to the stemcell VM, since it might cause disks on the stemcell to get cloned to the same UUID everywhere.

Having said that, I am not sure if that even makes sense--presumably the stemcell only has a system disk, and maybe we don't care even if that always has the same UUID. 

Anyway, for reference, here's what the vmx_options look like in VCenter for a working k8s worker node:
Screenshot 2025-08-13 at 11 15 51 AM

PR #402, opened 2025-09-09
Fix integration tests compatability with nsxt 3.2

Alphasite:
# Description

NSXT changed the return value for the list transport zone in 3.2, 4.2 and 9 by adding an additional valid value to the enum. It looks like the generated client validates the values returned by the server and thats blowing up when we make list call against newer nsx version.

So... why did we patch the 3.1.0 openapi spec instead of update to the 3.2.0 spec? Because that was a giant diff and we well create a seperate story for that.

## Related PR and Issues

[TNZ-55395]
[TNZ-54835]
[TNZ-56419]

Integration tests

## Type of change

## How Has This Been Tested?

- [ ] Manual verification 
- [ ] And integration tests passed in ci

# Checklist:


julian-hj (approved):
Looks OK to me. Bummer that we cannot just use a newer OpenAPIv3 spec, but getting a new release out is more important.

PR #414, opened 2025-10-24
fix vsan datastore renaming issue

julian-hj:
# Description

Previously when detaching a disk in a renamed vsan datastore, we erroneously located the underlying storage file for the disk, not the friendly symlinked location for the same file, and then attempted to move the former to the location of the latter, which failed since there was already a file in that location. This commit adds a check to skip the move operation when the destination file exists and the datastore is of type vsan. Also, various minor rubocop formatting fixes.
## Related PR and Issues
Fixes # 400
## Type of change

## How Has This Been Tested?

- Added units
- Tested manually in a bats environment with the cpi config swizzled to use the vsan datastore instead of nfs

# Checklist:

### DIFF
--- a/src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
+++ b/src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
@@ -41,11 +41,11 @@ module VSphereCloud
 
       def accessible_datastores
         Datastore.build_from_client(
-            @client,
-            host_properties['datastore'],
-            ensure_all: true
+          @client,
+          host_properties['datastore'],
+          ensure_all: true
         ).select(&:accessible)
-            .inject({}) do |acc, datastore|
+                 .inject({}) do |acc, datastore|
           acc[datastore.name] = datastore
           acc
         end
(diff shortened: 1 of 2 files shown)


KauzClay (approved):

src/vsphere_cpi/spec/unit/cloud/vsphere/resources/vm_spec.rb
```diff
@@ -420,11 +420,15 @@
   end
 
   describe '#detach_disks' do
+    let(:datastore_summary) { double(:datastore_summary, type: 'iscsi')}
```
Alphasite: nit: the whitespace is inconsistent here

src/vsphere_cpi/lib/cloud/vsphere/resources/vm.rb
```diff
@@ -389,22 +389,27 @@ def move_disks(disks_to_move)
...
             # Use current disk path incuded with the disk
-            dest_filename =  disk.backing.file_name.match(/^\[[^\]]+\] .*\/(.*)/)[1]
+            dest_filename = disk.backing.file_name.match(/^\[[^\]]+\] .*\/(.*)/)[1]
             dest_filename = "#{disk_path}/#{dest_filename}"
           end
           dest_path = "[#{current_datastore}] #{dest_filename}"
 
-          @client.move_disk(datacenter_mob, disk.backing.file_name, datacenter_mob, dest_path)
+          destination_exists = @client.disk_path_exists?(@mob, dest_path)
+          datastore_type = disk.backing.datastore.summary.type
```
Alphasite: Can we add a comment to explain the background for this and why we think this is safe? I think its reasonable but in 2 years im 100% sure i wont remember a word of it.

PR #420, opened 2025-11-05
Reaarange host group tests for more robustness

julian-hj:
# Description

Previously the tests for memory capacity used a host group and a resource pool with very similar capacity, which means that in VC8 the tests started failing because nsx-edge is using up some of the memory in the second cluster, changing the outcome of the VM placement. Switching the test around to use a resource pool that is considerably bigger should make the tests more predictable.

## Related PR and Issues
Fixes https://ci-demo.bosh-ecosystem.cf-app.com/teams/main/pipelines/vsphere-cpi/jobs/lifecycle-8.0-nsxt42-cvds/builds/30

none

## Type of change

test fix (no running code changed)

### DIFF
--- a/src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
+++ b/src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
@@ -465,14 +465,14 @@ describe 'Host Groups in Cluster and VM Host Rules' do
         context 'and one of the cluster has host group and other has resource pool' do
           let(:options) do
             cpi_options(
-                datacenters: [{
-                                  'datastore_pattern' => @datastore_shared_pattern,
-                                  'persistent_datastore_pattern' => @datastore_shared_pattern,
-                                  clusters: [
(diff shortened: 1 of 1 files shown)


src/vsphere_cpi/spec/integration/vm_host_rule_spec.rb
```diff
@@ -465,14 +465,14 @@
...
-                                  clusters: [
-                                      { @cluster_name => { host_group: {'name' => @second_host_group, 'drs_rule' => 'Must'} } },
-                                      { @second_cluster_name => { resource_pool: @second_cluster_resource_pool_name, } },
-                                  ]
-                              }]
+              datacenters: [{
+                              'datastore_pattern' => @datastore_shared_pattern,
+                              'persistent_datastore_pattern' => @datastore_shared_pattern,
+                              clusters: [
+                                { @cluster_name => { resource_pool: @first_cluster_second_resource_pool_name, } },
```
Alphasite: this is probably fine, my only worry is that its not exercising the nested resource group code path

PR #425, opened 2025-11-10
fix dns for bats tests

julian-hj:
Prevously we ran bosh create-env using the default 8.8.8.8 DNS server, but that dns server is no longer reachable from our testbeds. Use the local testbed DNS server instead.


Alphasite (approved):
🚢 it

PR #433, opened 2025-12-03
Account for system generated tag in tests

Alphasite:
# Description

VMs attached to a policy API managed segment get an additional tag on the port with the policy object ID. This should make the tests pass with this additional tag.


aramprice (approved):

aramprice:
Are there extra elements being found in these tests?

PR #464, opened 2026-05-28
Replace archived vsphere-automation-sdk with native TagClient

ay901246:
## Summary

The `vsphere-automation-sdk-ruby` gem was archived in 2021 and silently ignores `Configuration#ssl_ca_cert` — it never passes the bundle down to `Net::HTTP`, so peer verification always fails when a self-signed CA is configured. This broke `set_vm_metadata` for customers using custom CA certs (e.g. DATEV on VCF), which is why we had to revert in #463 and pin ops-manager back to v97.0.20.

This PR removes the dependency entirely and re-implements the handful of CIS tagging REST calls natively using the same `BaseHttpClient` that vCenter SOAP, NSX-T, and PBM already use — so CA pinning and skip-verify behavior are correct, consistent, and covered by unit tests.

## What changed

- **New** `lib/cloud/vsphere/tag_client.rb`: a lightweight client over `BaseHttpClient` covering all CIS tagging endpoints the CPI needs (`session`, `category`, `tag`, `tag-association`). Treats nil/blank/whitespace `ca_cert_file` as absent (preserving prior skip-verify behavior for deployments without a configured CA).
- **Rewritten** `lib/cloud/vsphere/attach_tag_to_vm.rb`: drops `require 'vsphere-automation-cis'` / `'vsphere-automation-vcenter'`. `InitializeConnection` now builds a `TagClient`. All public method signatures are unchanged so no callers needed updating.
- **Removed** `vsphere-automation-cis`, `vsphere-automation-vcenter`, `vsphere-automation-runtime` from `Gemfile`, `Gemfile.lock`, and `vendor/package/`.
- **New** `spec/unit/cloud/vsphere/tag_client_spec.rb`: 17 unit tests covering TLS/CA behavior (the live-HTTPS regression test that would have caught the v98.0.5 bug), all REST endpoint payload/header shapes, idempotent login, and error propagation.
- **Rewritten** `spec/unit/cloud/vsphere/attach_tag_to_vm_spec.rb`: stubs `TagClient` instead of SDK classes. Adds whitespace-only and nil `vcenter_connection_options` cases.
- **Updated** `spec/integration/attach_tag_to_vm_spec.rb`: drops SDK helper calls, adds a `BOSH_VSPHERE_CPI_CA_CERT_FILE`-gated integration case that verifies a real CIS session can be opened with a pinned CA (the direct regression test for TNZ-101747).

## Test plan

- [x] `bundle exec rspec spec/unit/cloud/vsphere/tag_client_spec.rb spec/unit/cloud/vsphere/attach_tag_to_vm_spec.rb` — 48 examples, 0 failures
- [x] `bundle exec rspec spec/unit` — 1020 examples, 0 failures
- [ ] CI pipeline (attach_tag integration suite with `BOSH_VSPHERE_CPI_CA_CERT_FILE` set against a VCF vCenter)

## Follow-up after merge

Once a new release is cut, unpin vsphere-cpi-release in ops-manager 3.1 and 3.3 (see TNZ-103786).

Made with [Cursor](https://cursor.com)


src/vsphere_cpi/lib/cloud/vsphere/attach_tag_to_vm.rb
```diff
@@ -91,63 +89,52 @@ def to_s
...
-      end
-
-      private def tag_association_api
-        @tag_association_api ||= VSphereAutomation::CIS::TaggingTagAssociationApi.new(@api_client)
-      end
+      # Builds a {TagClient} that connects to the configured vCenter and trusts
+      # the operator-provided CA bundle when one is present. The returned client
+      # is lazy: it does not contact vCenter until a method that requires a
+      # session is invoked.
+      def self.InitializeConnection(cloud_config, logger)
```
Alphasite: this structure feels wierd. I know its not something you introduced in your pr, but it may be worth seeing it you can opportunistically fix it since youre already working on it?

Alphasite: I think one thought i have is that either TagClient should be initialized entirely separately or totally internally, but here we have an odd flow where the factory method is on this class but its passed to this class' constructor

ay901246: Made the changes to have TagClient be initialized on its own via factory method.

julian-hj (approved):

